PR #12, opened 2020-10-15
fix postgres configration

ckildau:
remove ports definition for postgres instances in composer files. no need to proxy these ports externally, is available locally on the bridge.


sculptex:
Thanks Chris!

PR #25, opened 2021-03-03
Add github action for docker image build and test

satchinjoshi:
Blobber and validator docker images will be built and pushed to the docker hub when a new GitHub release is published.
The test will run before the image is built.

> The image tag will be the same as the tag that was used for the GitHub release

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,72 @@
+name: CI
+
+on:
+  [push]
+
+env:
+  BLOBBER_REGISTRY: ${{ secrets.BLOBBER_REGISTRY }}
+  VALIDATOR_REGISTRY: ${{ secrets.VALIDATOR_REGISTRY }}
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Install Go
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14.x
+    - uses: actions/checkout@v2
+    - name: Test
+      run: |
+        for mod_file in $(find * -name go.mod); do
```
kirillt, line 21: Let's create a small script `unit_test.sh`, so we will be able to run tests at GitHub, local machine or, potentially, GitLab later. With the script runs on any side will be the same, otherwise it will be easy to change unit tests logic in one place and forget to change another.

kirillt: Or `Makefile`, and we can put auxiliary Docker stuff there too (I don't think that we need releasing stuff itself in there, just stuff from `README`).

kirillt (commented):
Looks good, small change would be good

PR #35, opened 2021-03-22
Dependencies build updated

kirillt:
Right now, build fails with the following error:
```
Makefile:8: /tmp/bls/mcl/common.mk: No such file or directory
make: *** No rule to make target '/tmp/bls/mcl/common.mk'.  Stop.
```

This PR fixes it by downloading `bls` repo by git instead of `wget` and fetching submodules.


satchinjoshi (approved):
Seems to be working now. 
https://github.com/0chain/blobber/actions/runs/676260349

Issue #48, opened 2021-03-30
Security issue: File lists can be queried and collaborators added without authentication 


stewartie4:
Querying file and allocation metadata eg.
/v1/file/referencepath/
and
/v1/file/meta/
Requires no authentication other than the wallet Id and public key.

This may be a potential security problem as anyone can view which files are stored in an allocation and individual file metadata as long as they have that wallet's public key and id, all of which are readily available on the public transaction page when an allocation is created.
![image](https://user-images.githubusercontent.com/18306778/113047387-26170a00-9199-11eb-804a-baae887a49cd.png)

For example I queried a random allocation and could see it has 4 files, their names and when they were created
![image](https://user-images.githubusercontent.com/18306778/113047478-48a92300-9199-11eb-94d2-7c87ebe62d47.png)

These metadata requests (and others like it) should be BLS signed by the client. 
It would greatly increase security and would be transparent to the end user

stewartie4:
If someone can assign me to this I can start looking at it tomorrow/before the weekend

stewartie4:
Added the following endpoints to the scope of this issue:
/v1/file/objectpath
/v1/file/referencepath/list
/v1/file/stats
/v1/file/objecttree
/v1/file/collaborator **(This makes additional exploits possible, testing in progress)**
/v1/file/rename (however operation has to be committed which requires a signature)
/v1/file/copy (however operation has to be committed which requires a signature)
/v1/file/attributes (however operation has to be committed which requires a signature)
/v1/file/upload (however operation has to be committed which requires a signature)

MurashovVen:
> Added the following endpoints to the scope of this issue:
> /v1/file/objectpath
> /v1/file/referencepath/list
> /v1/file/stats
> /v1/file/objecttree
> /v1/file/collaborator **(This makes additional exploits possible, testing in progress)**
> /v1/file/rename (however operation has to be committed which requires a signature)
> /v1/file/copy (however operation has to be committed which requires a signature)
> /v1/file/attributes (however operation has to be committed which requires a signature)
> /v1/file/upload (however operation has to be committed which requires a signature)
> 
> **/v1/file/download is not directly susceptible to this but it can be made to initiate a file downloaded by adding a collaborator as above**
> 
[...]

PR #54, opened 2021-03-31
Add BLS signature to metadata endpoints

stewartie4:
PR to resolve https://github.com/0chain/blobber/issues/48 
Validation:
Valid signature - current behaviour
Invalid/missing signature: 
image

/v1/file/referencepath/
Signature is added as a query param eg.
image
/v1/file/meta/ 
Signature is added as part of the multipart form.

I debated adding the signature as a header to keep things consistent across endpoints so I would be grateful for some input with regards to that

SDK changes to allow the clients to support this new requirement
https://github.com/0chain/gosdk/pull/26

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ $ ./docker.local/bin/build.blobber.sh
 ```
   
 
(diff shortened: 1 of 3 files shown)


stewartie4:
closing due to increase in scope requiring an impl rethink

PR #63, opened 2021-04-03
Remove non working pull request check

Sriep:
The lint check on pull requests does not work. This needs to be removed so other people PR's can be merged.


satchinjoshi:
@peterlimg might be able to answer this as it was added based on his request.

Issue #81, opened 2021-04-18
Mock out gorm.DB to allow blobber unit tests


Sriep:
Currently it's not easy to write unit tests as blbber uses a raw gorm.DB in the code. It is generally bad to use active databases in unit tests, they should be mocked out. However, the way the blobber is written this is difficult to impossible.

The gorm.DB should be wrapped in an interface to make it easer to mock.

Sriep:
https://github.com/0chain/blobber/pull/87

shravanshetty1:
@Sriep What do you think about creating a store layer module.

This will make it easy to swap out underlying database at any time as long as the store interface has been implemented by the new database.
This will make it even easier to test as adding sqlmock is tedious.

PR #82, opened 2021-04-18
[WIP] Mock store

Sriep:
Issue https://github.com/0chain/blobber/issues/81
This warps the the *gorm.DB object in an interface `Transaction`. Also the state object has also been wrapped in an interface. 

This allows mock transactions to be used for unit tests and will make it easier to use different databases, if needed, in the future.

This has required changes in a lot of files. The key changes are in the `datastore` package, elsewhere they amount to
* `datastore.GetStore().GetTransaction(ctx)` changed to `datastore.GetTransaction(ctx* )`
* `datastore.GetStore().CreateTransaction(ctx)` changed to `datastore.CreateTransaction(ctx)`
* `.Error` changed to `.Error()`

To mock you can derive from MockTransaction and implicate mock code for any transaction methods you need in your test.

```go
type MyMockTransaction struct{
   datasotre.MockTransaction
}
func (t MyMockTransaction) Select(query interface{}, args ...interface{}) (tx Transaction) {
  ... some mock code
}


func TestSomething(t *testing.T) {
   datastore.MockTheStore(MockStore{})
   ... use Select somehow
}

```
Alternately if your unit test makes more use of the database, you can create a new `MyMockStore` object that implements the Store object with a mock database object, if the mock database object has the same methods as a `gorm.DB` then you can mostly cut and pastes on the `datastore.store` and `database.transaction` objects
```go
type myMockStore{db mocksqldb.DB}
func (store myMockStore) Open() error{
   ...
}
func (store *myMockStore) Close() {
   ...
}

func (store *myMockStore) CreateTransaction(ctx context.Context) context.Context {
   ...
}

func (store *myMockStore) GetTransaction(ctx context.Context) Transaction {
   ...
}
```
and somewhere in your unit test
```go
datastore.MockTheStore(MyMockStore)
```


Sriep:
I could replace the return type of Transaction methods with interface{} so that a *gorm.DB can be a Transaction object. Not sure if this helps or not.
```go
type Transaction interface {
	Model(value interface{}) (tx interface{})
...
	Order(value interface{}) (tx interface{})
	Create(value interface{}) (tx interface{})

}
```

Sriep:
Replaced with https://github.com/0chain/blobber/pull/87

PR #90, opened 2021-04-20
wip: Removing `RequestContext` from blobber.proto requests

### LINKED ISSUE #89: Redundant use of `RequestContext` in all methods in Blobber service
I think there is no need of using a separate `RequestContext` field in all the requests. We can easily allow certain headers using `ServeMuxOptions` and get them in the request context.

Ref: https://github.com/0chain/blobber/blob/master/code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto#L48

Shivam010:
Prototype change for removing RequestContext from request.

Closes #89


code/go/0chain.net/blobbercore/handler/helper.go
```diff
@@ -15,20 +18,57 @@ import (
...
-	ctx = context.WithValue(ctx, constants.CLIENT_KEY_CONTEXT_KEY,
-		r.ClientKey)
-	ctx = context.WithValue(ctx, constants.ALLOCATION_CONTEXT_KEY,
-		r.Allocation)
-	return ctx
+// setupGRPCHandlerContext will add client context keys in request context.Context
+// and return new context with client and client key values
+func setupGRPCHandlerContext(ctx context.Context, _allocation string) (context.Context, string, string) {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok || md == nil {
```
Sriep, line 25: Do you mean this? Or should this be `if !ok || md == nil`?

code/go/0chain.net/blobbercore/handler/helper.go
```diff
@@ -15,20 +18,57 @@ import (
...
+	}
+
+	client := getValue(common.ClientHeader)
+	clientKey := getValue(common.ClientKeyHeader)
+	// adding to context for future reference and calls
+	ctx = context.WithValue(ctx, constants.CLIENT_CONTEXT_KEY, client)
+	ctx = context.WithValue(ctx, constants.CLIENT_KEY_CONTEXT_KEY, clientKey)
+	ctx = context.WithValue(ctx, constants.ALLOCATION_CONTEXT_KEY, _allocation)
+	return ctx, client, clientKey
+}
```
Sriep, line 43: The two times that `setupGRPCHandlerContext` is used, the last parameter, `clientKey`, is not used. If it is unnessary, it should be removed.

Sriep (changes requested):
Is there a unit test that ensures that `setupGRPCHandlerContext` is correctly constructing the context?

code/go/0chain.net/blobbercore/handler/helper.go
```diff
@@ -15,20 +18,57 @@ import (
...
+
+func GrpcGatewayHeaderMatcher(key string) (string, bool) {
+	// check for default allowed headers
+	res, ok := runtime.DefaultHeaderMatcher(key)
+	if ok {
+		return res, ok
+	}
+	// allow header if one of our common header
+	allowed := map[string]struct{}{
+		common.ClientHeader: {},
```
shravanshetty1, line 53: Should probably be renamed to IsHeaderAllowed since thats what its doing. Other header matchers could mutate the header and prefix/suffix .etc.But this one just checks if the headers is allowed.

shravanshetty1 (changes requested):

PR #109, opened 2021-05-07
Check and prevent double registration for blobbers

andrenerd:


guruhubb:
@andrenerd @Sriep I think this check needs to be on the smart contract storagesc as well.

Sriep:
It would be easy enough to add a check to [addBlobber](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/blobber.go#L191). We have the new blobber and list of existing blobbers so should easy enough to prevent duplication here. Should 0chain replace and leave a warning or reject and return an error?

guruhubb:
@Sriep @andrenerd I think we should `replace` and leave a `warning`.

andrenerd:
@Sriep 
please, complete the review.

the discussion about further changes in storagesc is not supposed to prevent the pr from merging.
there will be another branch/pr for that changes in the 0chain repo.

shravanshetty1 (approved):
I think this is fine but it shouldnt be considered fixed until its handled on the smart contract side.
https://github.com/0chain/0chain/pull/272

PR #120, opened 2021-05-09
Add integration tests for the grpc endpoints

### LINKED ISSUE #88: Add integration tests for grpc api's

shravanshetty1:
closes #88


shravanshetty1:
@Sriep No longer wip

code/go/0chain.net/blobbercore/handler/grpc_handler_integration_test.go
```diff
@@ -0,0 +1,283 @@
...
+		config.Configuration.DBHost, config.Configuration.DBPort,
+		config.Configuration.DBUserName, config.Configuration.DBName,
+		config.Configuration.DBPassword)), &gorm.Config{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	tdController := NewTestDataController(db)
+
+	t.Run("TestGetAllocation", func(t *testing.T) {
+		err := tdController.ClearDatabase()
```
Sriep, line 79: Most new 0chain unit tests by @MurashovVen are written as [table-driven tests](https://blog.golang.org/subtests#TOC_2.). I think it would be a good idea to adopt that style here.

Also, can we have all tests running in parallel. 
```go
t.Run(...., ... {
    t.Parallel()
```

This is to help detect [race errors](https://golang.org/doc/articles/race_detector). To avoid confounding race errors run the unit test with the race flag, `go test -race` and remove race errors caused by the test and raise issues for race errors in the production code. Hint: try to copy by value to avoid unnecessarily sharing memory addresses between tests.

Sriep (changes requested):

code/go/0chain.net/blobbercore/handler/grpc_handler_integration_test.go
shravanshetty1: This is not a unit test - its an integration test. Its testing the api.

shravanshetty1: I have reformatted it to be table driven tests.

shravanshetty1: Handling race errors would be alot easier through unit tests, this is however integration tests.

If we were testing crud api, imagine the delete happening before the object was created. To solve this we will be forced to create large number of rows in tables such that its deleting something that already existed in db and wasnt created by api, it will be hard to tell which row is being used by which tests.

Since unit test does not even deal with db and doesnt care about db state it will be much easier to handle it there.

Sriep (approved):

Issue #137, opened 2021-05-18
Adding grpc equivalent of "/v1/file/attributes/{allocation}"


senpathi:
Migrate "/v1/file/attributes/{allocation}" UpdateAttributesHandler to GRPC

PR #152, opened 2021-05-26
137 grpc update object attributes handler

senpathi:
related to #137 

update object attributes handler migrated to grpc
tests added


code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -39,6 +39,21 @@ type AllocationChangeProcessor interface {
 	Unmarshal(string) error
 }
 
+type IAllocationChangeCollector interface {
+	TableName() string
+	AddChange(allocationChange *AllocationChange, changeProcessor AllocationChangeProcessor)
+	Save(ctx context.Context) error
```
shravanshetty1, line 45: A simpler solution would be just to create a function and put save and addchange inside the function. You can then mock that function instead of this

shravanshetty1 (changes requested):

code/go/0chain.net/blobbercore/allocation/allocationchange.go
senpathi: I can change this and add a function like AddChangeAndSave. But AddChange and Save functions were already there. I simply added them to the interface

shravanshetty1: Git diff said you created this interface - if you only added 2 functions it would show only those

PR #154, opened 2021-05-26
common changes added to grpc

senpathi:
common changes that are helpful to migrate object operation handlers


shravanshetty1:
not needed since its part of other PR's

Issue #158, opened 2021-05-26
Adding grpc equivalent of "/v1/file/download/{allocation}"


senpathi:
Migrate "/v1/file/download/{allocation}" DownloadHandler to GRPC

PR #160, opened 2021-05-26
158 grpc download file handler

senpathi:
relates to #158 

download file handler migrated to grpc
tests added


code/go/0chain.net/blobbercore/handler/object_operation_grpc_handler.go
```diff
@@ -0,0 +1,238 @@
...
+	"gorm.io/datatypes"
+	"gorm.io/gorm"
+	"strconv"
+)
+
+func (b *blobberGRPCService) DownloadFile(ctx context.Context, r *blobbergrpc.DownloadFileRequest) (
+	*blobbergrpc.DownloadFileResponse, error) {
+
+	allocationTx := r.Allocation
+	allocationObj, err := b.storageHandler.verifyAllocation(ctx, allocationTx, false)
```
shravanshetty1, line 22: Length of client id check has to be before this 
```

	if len(clientID) == 0 {
		return nil, common.NewError("download_file", "invalid client")
	}
```

shravanshetty1: I did not read this entire file - but can you reduce the changes between this function and original.

shravanshetty1 (changes requested):

code/go/0chain.net/blobbercore/mocks/FileStore.go
```diff
@@ -0,0 +1,275 @@
+// Code generated by mockery 2.7.5. DO NOT EDIT.
+
+package mocks
+
```
shravanshetty1, line 4: This file no longer needed

shravanshetty1 (changes requested):

PR #162, opened 2021-05-27
Upgrade allocation payer-based logic

andrenerd:


Sriep (changes requested):
They automatic checks should prevent syntax errors but don't. Issue https://github.com/0chain/blobber/issues/195

We need unit tests for modified methods such as `blobberGRPCService.GetFileMetaData` and `StorageHandler.DownloadFile`.  These tests have not been written yet, but it would add confidence to write the tests here.

Sriep:
Merge in the latest master changes, as I think that might fix the failing unit test check.

PR #180, opened 2021-06-02
Workaround for double registration for restarted blobbers

### LINKED ISSUE #182: Failed during registering blobber to the mining network, it's duplicated
```
2021-06-02T12:08:29.436Z	INFO	handler/protocol.go:85	Failed during registering blobber to the mining network, it's duplicated
2021-06-02T12:08:59.437Z	ERROR	blobber/main.go:414	Add blobber transaction could not be verified	{"err": "Duplicated", "txn.Hash": ""}
main.RegisterBlobber
	/blobber/go/0chain.net/blobber/main.go:414
```

andrenerd:
To return valid hash, not an error.
To prevent issues during re-registration requests from restarted blobbers.

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/protocol.go
+++ b/code/go/0chain.net/blobbercore/handler/protocol.go
@@ -82,8 +82,8 @@ func RegisterBlobber(ctx context.Context) (string, error) {
 
 	for _, sRegisteredNode := range sRegisteredNodes {
 		if sn.ID == string(sRegisteredNode.ID) || sn.BaseURL == sRegisteredNode.BaseURL {
-			Logger.Info("Failed during registering blobber to the mining network, it's duplicated")
-			// should return valid transcation hash, notan error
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/blobbercore/handler/protocol.go
```diff
@@ -83,7 +83,9 @@ func RegisterBlobber(ctx context.Context) (string, error) {
 	for _, sRegisteredNode := range sRegisteredNodes {
 		if sn.ID == string(sRegisteredNode.ID) || sn.BaseURL == sRegisteredNode.BaseURL {
 			Logger.Info("Failed during registering blobber to the mining network, it's duplicated")
```
bbist, line 85: Looks like this workaround might stick around. Maybe it'd make sense to update the log to reflect what's happening after the change. Something like, "blobber already registered, updating settings".

bbist (commented):
Looks good. I assume this is a workaround.

PR #190, opened 2021-06-05
createDir function

moldis:
Possibility to create dir without files inside

`zbox createdir --allocation dd60f818349bac486f95780a2024c3c17e6e4fa598cd324206381d4cd6733f85 --dirname /test777`

`zbox list-all --allocation dd60f818349bac486f95780a2024c3c17e6e4fa598cd324206381d4cd6733f85`

Will include new dir

`"name":"test777","num_of_blocks":2,"path":"/test777","path_hash":"a7ffc6f8bf1ed76651c14756a061d662f580ff4de43b49fa82d80a4b80f8434a","size":2,"type":"d","updated_at":"2021-07-12T15:06:05.419607Z"}`


lpoli:
For file upload blobber first puts the file in temporary directory for certain time and only upon commit transaction the file is moved to permanent directory.
CreateDir function has put new directory path directly into permanent directory.
I suppose Blobber should get paid for empty directory too.

stewartie4:
This is great! A much needed feature 

If I understand right though, this action does not need committed? 

In terms of payment, an empty directory technically takes up some space in the file system table but ultimately is empty - should blobbers be paid for an empty directory? 
Unless a temp hidden file is committed to an empty directory thereby enabling the blobber to be paid

guruhubb:
Blobbers do not charge for rename, share, delete, copy, move, add-collab, and emptyDir creation.  Blobbers do carry the cost of commit transactions to the chain but we expect that to be minimal. If there is an crud attack, the Blobbers can reduce the commit frequency.

Sriep (approved):

code/go/0chain.net/blobbercore/allocation/newfilechange.go
```diff
@@ -43,9 +43,79 @@ type NewFileChange struct {
...
+	}
+
+	var newDir = reference.NewDirectoryRef()
+	newDir.ActualFileSize = 2
+	newDir.AllocationID = dirRef.AllocationID
+	newDir.MerkleRoot = nf.MerkleRoot
+	newDir.Name = dirName
+	newDir.Size = 2
+	newDir.NumBlocks = 2
+	newDir.ParentPath = dirRef.Path
```
cnlangzi, line 95: I think it should be 
```
newDir.ActualFileSize = 0
newDir.Size = 0
newDir.NumBlocks = 0
```

Issue #178, opened 2021-06-01
increase sql table owner_public_key from varchar(256) to 512


mallochine:
The MIRACL public key format has 258 characters  (example: 0420ec9dd58f2d16babe5eddd0990472ef7f05fd707e3a4171e61ba9360fe4a3891d9bcf8a86b0b270d104c1a2a093f744d27793927910b38be9883ec0b06ca3490569516d05cfdfb66f64ac13681905ec639930dc66ae889944d4a0e3fbf249fc0d52adb860e55487251acbd90e8f1b5563e71e67af1e110ce986ff41df34a7ed)

very sadly, this 258 string is just 2 character more than the 256 limit.

The error message I'm seeing is:

`{"code":"invalid_parameters","error":"invalid_parameters: Invalid allocation id passed.verify_allocation: verifying allocation transaction error: ERROR: value too long for type character varying(256) (SQLSTATE 22001)"}`

It can be replicated with this curl call (although the six network is currently down unfortunately)
```
curl 'http://six.devnet-0chain.net:31301/v1/file/upload/d2f7e87dcd5dce7431112310ca811435925a40f1758c51cc7d3d9a93bfd3cdf1' \
  -H 'Connection: keep-alive' \
  -H 'Pragma: no-cache' \
  -H 'Cache-Control: no-cache' \
  -H 'x-app-client-id: 69c4e00f68457c76a473e71368effd6542853c61d27f9bb1136bc131d96b428d' \
  -H 'x-app-client-key: 0420ec9dd58f2d16babe5eddd0990472ef7f05fd707e3a4171e61ba9360fe4a3891d9bcf8a86b0b270d104c1a2a093f744d27793927910b38be9883ec0b06ca3490569516d05cfdfb66f64ac13681905ec639930dc66ae889944d4a0e3fbf249fc0d52adb860e55487251acbd90e8f1b5563e71e67af1e110ce986ff41df34a7ed' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36' \
  -H 'content-type: multipart/form-data; boundary=6834d1dc50f4e20a1997f3342a4495650798577a0421bb1ace2877d26c67' \
  -H 'Accept: */*' \
  -H 'Origin: http://localhost:82' \
  -H 'Referer: http://localhost:82/' \
[...]

PR #191, opened 2021-06-07
increase write_markers.client_key size

mallochine:
still seeing issue #178 , client_key is also a public key even though it doesn't have the "pub" keyword in it


mallochine:
@shravanshetty1 ok I've tested this by going on six network's rancher, going to each postgres, executing that alter sql, and now the file upload is making progress again.

andrenerd (approved):

PR #200, opened 2021-06-09
for extra precaution, 2nd conversion of public key from MIRACL to herumi

### LINKED ISSUE #198: Need to handle MIRACL signature
as the wasm file sends a MIRACL signature in "X-App-Client-Signature" header, we need to convert this over to herumi signature.

mallochine:
ok this is tested on six network and successfully moves past the first stage of uploading blobber.

merging this in will resolve #198

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/storage_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/storage_handler.go
@@ -712,6 +712,7 @@ func verifySignatureFromRequest(r *http.Request, pbK string) (bool, error) {
 	}
 
 	hash := encryption.Hash(data)
+	pbK = encryption.MiraclToHerumiPK(pbK)
 	return encryption.Verify(pbK, sign, hash)
 }
 


code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -712,6 +712,7 @@ func verifySignatureFromRequest(r *http.Request, pbK string) (bool, error) {
 	}
 
 	hash := encryption.Hash(data)
+	pbK = encryption.MiraclToHerumiPK(pbK)
```
andrenerd, line 715: Feel free to merge if you want,
but i personally wouldn't do that, assigning value with changed "essense" to the same var.
maybe to call it pbKHerumi... or not creating it at all and just wrapping the passed param with encryption.MiraclToHerumiPK().

andrenerd (changes requested):
I don't insist on the change. Up to you.

mallochine:
commit already merged in another branch.

PR #214, opened 2021-06-15
Adding grpc endpoints

shravanshetty1:
This is an umbrella issue for adding grpc endpoints to blobber.

How can we verify this isnt causing breaking changes?
There are integration tests and it will also be deployed on one of the devnets for a few days to fix any minor bugs


shravanshetty1:
@Sriep @andrenerd Please have a brief look at this to review general decisions, a detailed review is not required since there are tests already and its also being used in nine.devnet with no issues so far.

andrenerd (approved):

PR #246, opened 2021-07-15
WIP: Remove the attribute filed which is not used anywhere.

Sriep:
Part of https://github.com/0chain/zboxcli/issues/58.


cnlangzi (changes requested):

Sriep:
We don't need this any more

Issue #43, opened 2021-03-30
Use 'buf' to run grpc commands


shravanshetty1:

shravanshetty1:
https://github.com/bufbuild/buf

meenal06:
Ping @shravanshetty1 ,

kushthedude:
Ping @shravanshetty1 , I guess I dont have triage rights thats why I cant assign issues to myself.

PR #250, opened 2021-07-16
feat: initialise buf congiurations & generate yamls

kushthedude:
reference #43

### DIFF
--- a/code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
+++ b/code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
@@ -66,7 +66,6 @@ service Blobber {
       }
       additional_bindings: {
         delete: "/v2/file/upload/{allocation}"
-        body: "*"
       }
     };
   }
--- a/code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
+++ b/code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
@@ -66,6 +66,7 @@ service Blobber {
       }
       additional_bindings: {
         delete: "/v2/file/upload/{allocation}"
+        body: "*"
       }
     };
   }
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -1,0 +1,11 @@
+version: v1beta1
+plugins:
+  - name: go
+    out: proto
+    opt: paths=source_relative
+  - name: go-grpc
+    out: proto
+    opt: paths=source_relative,require_unimplemented_servers=false
+  - name: grpc-gateway
+    out: proto
+    opt: paths=source_relative
--- a/buf.yaml
+++ b/buf.yaml
@@ -1,0 +1,18 @@
+version: v1beta1
+name: buf.build/0chain/blobber
+deps:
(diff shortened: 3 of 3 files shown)


kushthedude:
Introspected Roadmap:- 

- Initialise buf.
- Split prots on the basis of services & contract
- Organise the generated stubs into what buf expects.
- Remove the redundant CI steps related to grpc & introduce buf in grpc pipelines.

cc: @senpathi @guruhubb @shravanshetty1

kushthedude:
Please ignore the generated-failure right now, I didn't regenerate the stubs as we will be deprecating protoc and move out to buf build

code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
```diff
@@ -66,7 +66,6 @@ service Blobber {
       }
       additional_bindings: {
         delete: "/v2/file/upload/{allocation}"
-        body: "*"
       }
```
shravanshetty1, line 69: There is a delete api that accepts data in its body, you should handle this in a separate issue

shravanshetty1 (changes requested):
@kushthedude you also need to update the documentation in the blobbercore/blobbergrpc folder

code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber.proto
kushthedude: It is not possible, DELETE request body is explicitly discarded by the HTTP protocol.

kushthedude: FYI. where it is written, please read this :
```
The spec does not explicitly forbid or discourage it, so I would tend to say it is allowed.

Microsoft sees it the same way (I can hear murmuring in the audience), they state in the MSDN article about the DELETE Method of ADO.NET Data Services Framework:

If a DELETE request includes an entity body, the body is ignored [...]
Additionally here is what RFC2616 (HTTP 1.1) has to say in regard to requests:

an entity-body is only present when a message-body is present (section 7.2)
the presence of a message-body is signaled by the inclusion of a Content-Length or Transfer-Encoding header (section 4.3)
a message-body must not be included when the specification of the request method does not allow sending an entity-body (section 4.3)
an entity-body is explicitly forbidden in TRACE requests only, all other request types are unrestricted (section 9, and 9.8 specifically)
For responses, this has been defined:

whether a message-body is included depends on both request method and response status (section 4.3)
a message-body is explicitly forbidden in responses to HEAD requests (section 9, and 9.4 specifically)
a message-body is explicitly forbidden in 1xx (informational), 204 (no content), and 304 (not modified) responses (section 4.3)
all other responses include a message-body, though it may be of zero length (section 4.3)
```

However, not going into much details this issue should be separately handled.

kushthedude:
> @kushthedude you also need to update the documentation in the blobbercore/blobbergrpc folder

This will be done when the integration with buf is complete, there is a lot of semantic as well as definition error in the protos including the package versioning, best practices & generated stubs. Once it is done & buf is integrated into our CI pipelines then I think it would be better to remove the protoc with buf in the repository.

PR #272, opened 2021-07-28
hotfix: check if the http hadnlers were misbehaving too

kushthedude:
Please dont merge this PR, it is to check if the grpc tests needs some love or the logic implemented is just as is.


kushthedude:
Fixedd

PR #273, opened 2021-07-28
File References Pagination

lpoli:
[ISSUE](https://github.com/0chain/gosdk/issues/117)

This branch solves the above issues by paginating File References which is used to construct ObjectTree, ObjectPath and ReferencePath. I've created new handler that provides paginated fileRefs.
There are three types of fileRefs: regular(undeleted refs), updated(updated refs compared to some given time) and deleted(deleted refs compared to some given time) which is crucial to construct local Tree cache to Add, Update and Delete tree nodes.

Pagination is done using alternative of offset https://blog.jooq.org/2013/10/26/faster-sql-paging-with-jooq-using-the-seek-method/
This makes use of indexes which significantly improves search performance.

Default and maximum pageLimit is 100 rows(about 50KB to 100KB json response). Client can request to obtain single reference with pageLimit 1. GoSdk work is in progress which serves as validating consensus of fileRefs. So that means actual tree construction is upto the application that is build upon gosdk; For example, zboxCli, 0fs, s3fs, ipfs, etc.


code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -239,6 +239,20 @@ func ObjectTreeHandler(ctx context.Context, r *http.Request) (interface{}, error
 	return response, nil
 }
 
+func RefsHandler(ctx context.Context, r *http.Request) (interface{}, error) {
+	if r.Method == "POST" {
+		return nil, common.NewError("invalid method", "Invalid method used. Use GET instead")
```
kushthedude, line 244: Have this check in the router itself, gorilla is capable of specifyng request type with the handlers

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -56,40 +56,40 @@ func (a *Attributes) Validate() (err error) {
...
-	Children            []*Ref         `gorm:"-"`
-	childrenLoaded      bool
-
-	OnCloud        bool            `gorm:"column:on_cloud" filelist:"on_cloud"`
-	CommitMetaTxns []CommitMetaTxn `gorm:"foreignkey:ref_id" filelist:"commit_meta_txns"`
-	CreatedAt      time.Time       `gorm:"column:created_at" dirlist:"created_at" filelist:"created_at"`
-	UpdatedAt      time.Time       `gorm:"column:updated_at" dirlist:"updated_at" filelist:"updated_at"`
-
-	DeletedAt gorm.DeletedAt `gorm:"column:deleted_at"` // soft deletion
+	ID                  int64          `gorm:"column:id;primary_key" json:"id,omitempty"`
```
kushthedude, line 59: Why add json annotations explicitly? Gorm handles the json naming imo.

kushthedude (changes requested):
Following is a review at first glance, will take a deeper look at the logic tonight. 
Can you also add tests for the pagination?

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -26,6 +26,10 @@ const (
 
 	DOWNLOAD_CONTENT_FULL  = "full"
 	DOWNLOAD_CONTENT_THUMB = "thumbnail"
+	PAGE_LIMIT             = 100 //100 rows will make upto 100 KB
```
kushthedude, line 29: Use Camel Case naming

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -26,6 +26,10 @@ const (
...
+	UPDATED                = "updated"
+	DELETED                = "deleted"
+	REGULAR                = "regular"
```
kushthedude, line 32: What is the need for these constants, we can directly use the string name in the switch?

kushthedude (commented):
I don't have context on 0fs and the pagination followed here, according to the original issue the need was to have `/v1/file/list/{allocation}` accept query params & support pagination. Since I am still new to the project, I have just pointed out the general golang improvements.

code/go/0chain.net/blobbercore/handler/handler.go
lpoli: Done

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: Done

code/go/0chain.net/blobbercore/reference/ref.go
lpoli: It gives flexibility and also I am not sure about Gorm handling json naming. It supports json dataTypes but it always fills struct fields with its value and its easy to use it with json tag.

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: > I don't have context on 0fs and the pagination followed here, according to the original issue the need was to have `/v1/file/list/{allocation}` accept query params & support pagination. Since I am still new to the project, I have just pointed out the general golang improvements.

Its actually replacement to `/v1/file/objectpath/{allocation}`, ` /v1/file/referencepath/{allocation}` and  `/v1/file/objecttree/{allocation}`
But it can be used widely by any use cases where it is required to fetch large trees.

This allows client application to build any kind/size of tree rather than overloading blobber.

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: Done

lpoli:
> Following is a review at first glance, will take a deeper look at the logic tonight.
> Can you also add tests for the pagination?

Can I add it later after completing 0FS?
Should not take long

kushthedude (approved):

kushthedude:
@lpoli resolve the conflicts, else lgtm

code/go/0chain.net/blobbercore/reference/ref.go
kushthedude: json column field is used when we unmarshal an interface to the following dataType, but since we already have gorm annotation, gorm has default serialising strategy as `snake_case` hence I said this change is redundant.

kushthedude (commented):
Hey @lpoli, CI is complaining will you please check into it, also if you like let's make this change too, https://github.com/0chain/blobber/pull/273#discussion_r679342205 either is fine with me 😅

lpoli:
> Hey @lpoli, CI is complaining will you please check into it, also if you like let's make this change too, [#273 (comment)](https://github.com/0chain/blobber/pull/273#discussion_r679342205) either is fine with me sweat_smile

CI test is error is fixed.
About json tag I say lets use json tag as I could not find explicit method that gorm provides to convert row data to json data. Also I checked if json.Marshall/Unmarshall could use other tag(gorm in this case) but in vain.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -56,40 +56,40 @@ func (a *Attributes) Validate() (err error) {
...
+	Type                string         `gorm:"column:type" dirlist:"type" filelist:"type" json:"type,omitempty"`
```
cnlangzi, line 60: please revert changes on `json` tag. It might break much old codes. (eg all api client, and all json unmarshal...etc)

code/go/0chain.net/blobbercore/blobberhttp/response.go
```diff
@@ -34,6 +34,14 @@ type ReferencePathResult struct {
 	LatestWM *writemarker.WriteMarker `json:"latest_write_marker"`
 }
 
+type RefResult struct {
+	TotalPages    int                      `json:"total_pages"`
+	NewOffsetPath string                   `json:"offsetPath,omitempty"`
```
cnlangzi, line 39: could you add some documentation about what is NewOffsetPath/NewoffsetDate?  it is confused there without any comment.

code/go/0chain.net/blobbercore/handler/helper.go
```diff
@@ -0,0 +1,17 @@
+package handler
+
+import (
+	"time"
+
+	"github.com/0chain/blobber/code/go/0chain.net/core/common"
+)
+
+func checkValidDate(s string) error {
```
cnlangzi, line 9: how about Timezone? it will be an issue if client send date with local time zone.

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -667,6 +669,118 @@ func (fsh *StorageHandler) GetObjectTree(ctx context.Context, r *http.Request) (
...
+	}
+
+	refType := r.FormValue("refType")
+	var refs *[]reference.Ref
+	var totalPages int
+	var newOffsetPath string
+	var newOffsetDate string
+
+	switch {
+	case refType == "regular":
```
cnlangzi, line 746: do you think it is better to implement them with restful style? regular = GET/updated=PUT/deleted=DELETE.

cnlangzi: I see them on `reference`. They mean `order by`. hmm, the naming is a bit confusing.

cnlangzi: do you think it is better to rename them 
- GetRefsOrderByUpdated
- GetRefsOrderByDeleted

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -99,3 +102,178 @@ func GetObjectTree(ctx context.Context, allocationID string, path string) (*Ref,
...
+	}
+	totalPages = int(math.Ceil(float64(totalRows) / float64(pageLimit)))
+	return
+}
+
+//Retrieves updated refs compared to some update_at value. Useful to localCache
+func GetUpdatedRefs(ctx context.Context, allocationID, path, offsetPath, _type, updatedDate, offsetDate string, level, pageLimit int) (refs *[]Ref, totalPages int, newOffsetPath, newOffsetDate string, err error) {
+	var totalRows int64
+	var pRefs []Ref
+	db := datastore.GetStore().GetDB()
```
cnlangzi, line 167: Yes. code smells off. but i have not idea yet.

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: All the endpoints are actually GET request. We Get;
Regular refs: Those refs that are not deleted
Updated refs: Those refs that are updated compared to the given time reference
Deleted refs: Those refs that are deleted compared to the given time reference

So it doesn't make sense to use PUT and DELETE since they mean to update and delete the rows but here we only get the rows.

cnlangzi: got it.

code/go/0chain.net/blobbercore/blobberhttp/response.go
cnlangzi: `total_pages` is snake_case, and offsetPath is lower_case. should we use snake_case only?

cnlangzi: and the field name is NewOffsetPath, but it is `offsetPath` in json. why they are different?

cnlangzi:
revert json tag on https://github.com/0chain/blobber/pull/325

PR #274, opened 2021-07-29
Changing gosdk to v0.0.0

moldis:


kushthedude (changes requested):
We don't release 0.0.0

moldis:
> We don't release 0.0.0

Since last gosdk release, it will be released together with actuall release, i.e. v0.0.0 and v1... etc.

kushthedude:
> Since last gosdk release, it will be released together with actuall release, i.e. v0.0.0 and v1... etc.

That;s not how releases works, refer this please https://semver.org/

kushthedude (changes requested):
This should be discussed before merging, lets not add more tech-debt to the project. We already have a lot to tackle once we are on mainnet.

PR #276, opened 2021-07-29
fix: go.mod & .sum conflicts due 0.0.0

kushthedude:


kushthedude:
Waiting for build to pass, since it is a trivial change with removing some unused hashes from go.sum. I will be merging it if CI passes without any issues

Issue #249, opened 2021-07-16
CI: Move to GitHub Actions providers


kushthedude:
## Problem

We are using a lot of bare metal packages like fetching golangci-lint & other utils by scripts or go to fetch deps, rather a lot of standard gh-actions are available for all this like golangci-lint & code-cov.
One of the major features of why people use gh-actions is due to the available actions in the marketplace, by not using them we are voiding & missing out the major characterstic.

## Solution:
Move out to the available github-actions for lint, test & other utils.

PR #280, opened 2021-07-30
feat: introduce golangci-lint action for faster builds

kushthedude:

- If anyone has issues with moving to official gh-action rather than manually fetching & setting up deps, see the performance spec https://github.com/golangci/golangci-lint-action#performance (it is around 35X faster)!!!!!!
- Go version in the modules setup was 1.13 however in the CI, it is specified as 1.14.x which is ambiguous fixed that too.

fixes https://github.com/0chain/blobber/issues/249

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,13 +20,9 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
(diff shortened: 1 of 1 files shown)


kushthedude:
1minute 30 seconds in older PRs, 10 seconds only in this PR, at least 10X fast over in just the first run, here.https://github.com/0chain/blobber/runs/3192199147 !

PR #281, opened 2021-08-02
feat: revamp GetReferencePath to grpc

kushthedude:


kushthedude:
Tests passed, HTTP tests commented out in case of rollback. Things look good, revamping the GetReferencePath is a success, taking the audacity to merge the PR.

PR #283, opened 2021-08-02
feat: revamp calculatehash to use internal grpc communication 

kushthedude:


kushthedude:
Tests passed, HTTP tests as well as handlers logic is commented out not removed, just in case of rollback. Things look good, revamping the Calculatehash is a success, taking the audacity to merge the PR.

PR #284, opened 2021-08-03
feat: revamp commitmetatxn to use grpc internal comm

kushthedude:


kushthedude:
Tests passed, HTTP tests as well as handlers logic is commented out not removed, just in case of rollback. Things look good, revamping the commitmetatxn is a success, taking the audacity to merge the PR.

PS: Fixing tests is a lot of pain 😢

PR #287, opened 2021-08-04
 feat: revamp UpdateObjectAttribute to use grpc for internal communication

kushthedude:


kushthedude:
Tests passed, HTTP tests, as well as handlers logic, is commented out not removed, just in case of rollback. Things look good, revamping the UpdateObjectAttribute is a success, taking the audacity to merge the PR.

PS: Fixing tests is a lot of pain 😓

PR #291, opened 2021-08-04
feat: revamp CopyObject handler to follow grpc internal comm 

kushthedude:


kushthedude:
Tests passed, HTTP tests, as well as handlers logic, is commented out not removed, just in case of rollback. Things look good, revamping the CopyObject is a success, taking the audacity to merge the PR.

Anyone having any issue/doubts about this PR, please read about grpc-gateway, errors wrapping.

PS: Making error catching was a tedious task here & moreover with revamping the redundant logic to one was another.

PR #295, opened 2021-08-05
fix(lock):#294 fixed memory leak issue

### LINKED ISSUE #294: It might be a memory leak issue
https://github.com/0chain/blobber/blob/de66db8d0ec6bc380b83d7b47e2995f10acdaac3/code/go/0chain.net/core/lock/lock.go#L8

sync.Mutex is always appended to lockPool. but it is never removed. It might be a memory leak issue.

```
package lock

import "sync"

var lockPool = make(map[string]*sync.Mutex, 0) //nolint:gosimple // need more time to verify: declaring capacity is probably necessary?
var lockMutex = &sync.Mutex{}

func GetMutex(tablename string, key string) *sync.Mutex {
	lockKey := tablename + ":" + key
	lockMutex.Lock()
	defer lockMutex.Unlock()
	if eLock, ok := lockPool[lockKey]; ok {
		return eLock
	}
	lockPool[lockKey] = &sync.Mutex{}
	return lockPool[lockKey]
}
```

cnlangzi:
fixed #294

### DIFF
--- a/code/go/0chain.net/core/lock/lock.go
+++ b/code/go/0chain.net/core/lock/lock.go
@@ -1,17 +1,70 @@
 package lock
 
-import "sync"
+import (
+	"sync"
+	"time"
+)
 
-var lockPool = make(map[string]*sync.Mutex, 0) //nolint:gosimple // need more time to verify: declaring capacity is probably necessary?
-var lockMutex = &sync.Mutex{}
+var (
+	lockPool   = make(map[string]*Mutex)
+	unlockPool = make(map[string]time.Time)
+	lockMutex  sync.Mutex
+)
 
-func GetMutex(tablename string, key string) *sync.Mutex {
+// Mutex a mutual exclusion lock.
+type Mutex struct {
+	// key lock key in pool
+	key string
+	sync.Mutex
+}
+
+// Lock implements Locker.Lock
+func (m *Mutex) Lock() {
+	m.Mutex.Lock()
+}
+
+// Unlock implements Locker.Unlock, and mark mutex as unlock object
+func (m *Mutex) Unlock() {
+	lockMutex.Lock()
+	defer lockMutex.Unlock()
+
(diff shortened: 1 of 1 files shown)


moldis:
Sounds good. Could you give some trace before your changes (with memory leak) and after, to be sure it's gone?

code/go/0chain.net/core/lock/lock.go
```diff
@@ -1,17 +1,70 @@
...
 
-import "sync"
+import (
+	"sync"
+	"time"
+)
 
-var lockPool = make(map[string]*sync.Mutex, 0) //nolint:gosimple // need more time to verify: declaring capacity is probably necessary?
-var lockMutex = &sync.Mutex{}
+var (
```
Sriep, line 8: It might be better to wrap these global objects inside a mutex factory object.

code/go/0chain.net/core/lock/lock.go
```diff
@@ -1,17 +1,70 @@
...
+	lockPool[lockKey] = m
+
+	return m
+}
+
+func init() {
+	go startLockPoolCleaner()
+}
+
+func startLockPoolCleaner() {
```
Sriep, line 57: Personally, I would prefer to use a reference counting method, you could use [SetFinalizer](https://pkg.go.dev/runtime#SetFinalizer) to control the count.

However, if using this method I think it might be best to control the loop from blobber.main, perhaps add to the other [worker threads](https://github.com/0chain/blobber/blob/master/code/go/0chain.net/blobber/main.go#L179).

Sriep (changes requested):
This should have a unit test.

cnlangzi:
> Sounds good. Could you give some trace before your changes (with memory leak) and after, to be sure it's gone?

I have updated detail about memory leak on https://github.com/0chain/blobber/issues/294.

PR #297, opened 2021-08-05
Fix with revoked share

moldis:
--revoke share didn't work since the last gosdk update. This fix should resolve it

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/handler.go
+++ b/code/go/0chain.net/blobbercore/handler/handler.go
@@ -455,7 +455,7 @@ func InsertShare(ctx context.Context, r *http.Request) (interface{}, error) {
 
 	existingShare, _ := reference.GetShareInfo(ctx, authTicket.ClientID, authTicket.FilePathHash)
 
-	if existingShare != nil {
+	if existingShare != nil && len(existingShare.OwnerID) > 0 {
 		err = reference.UpdateShareInfo(ctx, shareInfo)
 	} else {
 		err = reference.AddShareInfo(ctx, shareInfo)
--- a/code/go/0chain.net/blobbercore/handler/handler.go
+++ b/code/go/0chain.net/blobbercore/handler/handler.go
@@ -453,10 +453,7 @@ func InsertShare(ctx context.Context, r *http.Request) (interface{}, error) {
 		ExpiryAt:                  common.ToTime(authTicket.Expiration),
 	}
 
-	existingShare, err := reference.GetShareInfo(ctx, authTicket.ClientID, authTicket.FilePathHash)
-	if err != nil {
-		return nil, err
-	}
+	existingShare, _ := reference.GetShareInfo(ctx, authTicket.ClientID, authTicket.FilePathHash)
 
 	if existingShare != nil {
 		err = reference.UpdateShareInfo(ctx, shareInfo)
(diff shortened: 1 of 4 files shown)


code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -453,12 +453,9 @@ func InsertShare(ctx context.Context, r *http.Request) (interface{}, error) {
 		ExpiryAt:                  common.ToTime(authTicket.Expiration),
 	}
 
-	existingShare, err := reference.GetShareInfo(ctx, authTicket.ClientID, authTicket.FilePathHash)
-	if err != nil {
-		return nil, err
-	}
```
cnlangzi, line 459: should  we check err here?  We need to know if db throw any error?

```
func GetShareInfo(ctx context.Context, clientID string, filePathHash string) (*ShareInfo, error) {
	shareInfo := &ShareInfo{}
	err := db.Table(TableName()).
		Where(&ShareInfo{
			ClientID:     clientID,
			FilePathHash: filePathHash,
		}).
		First(shareInfo).Error

	if err == gorm.ErrRecordNotFound {
		return nil, nil
	}
	if err != nil {
		return nil, err
	}
	return shareInfo, nil
}

```

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -423,19 +423,26 @@ func (fsh *StorageHandler) DownloadFile(
...
+		shareInfo, err = reference.GetShareInfo(
 			ctx,
 			readMarker.ClientID,
 			authToken.FilePathHash,
 		)
-		if err != nil {
-			return nil, errors.New("error during share info lookup in database" + err.Error())
-		} else if shareInfo == nil || shareInfo.Revoked {
+
+		if err == nil && shareInfo.Revoked {
```
cnlangzi, line 434: app might crash if shareInfo is nil?  It is better to check it first.

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/handler/handler.go
moldis: Nope, it's wrong 'tricky' implementation that causes previous bugs. I removed this check from DB level

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
moldis: This is exactly what was done in the previous comment. It' can't be nil if err is nil.

cnlangzi:
here is an example for my review. thanks
```
var ErrServiceUnavailable = errors.New("service unavailable")
if err != nil { //any network/db error
    if errors.Is(err, ErrHasNotShared) {
         return nil, errors.New("client does not have permission to download the file. share revoked")
    }
    traceid := logging.Error(err)
    return nil, errors.Thrown(ErrServiceUnavailable,traceid)
}

if shareInfo == nil ||  shareInfo.Removed {
    return nil, errors.New("client does not have permission to download the file. share revoked")
}


var ErrHasNotShared = errors.New("file is not shared")


func GetShareInfo(ctx context.Context, clientID string, filePathHash string) (*ShareInfo, error) {
	shareInfo := &ShareInfo{}
	err := db.Table(TableName()).
		Where(&ShareInfo{
			ClientID:     clientID,
			FilePathHash: filePathHash,
		}).
		First(shareInfo).Error

	if err != nil {

    if errors.Is(err, gorm.ErrRecordNotFound) {
        return nil, ErrHasNotShared
    }

		return nil, err
	}
	return shareInfo, nil
}
```

cnlangzi:
BTW, `errors.Thrown` is in `github.com/0chain/errors` package.

moldis:
@cnlangzi This has to be applied globally and it's not following my current logic for sharing

PR #314, opened 2021-08-23
Download file handler grpc gateway WIP

### LINKED ISSUE #313: upload/download http handlers to grpc middleware

senpathi:
resolves #313

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/grpc_helpers.go
+++ b/code/go/0chain.net/blobbercore/handler/grpc_helpers.go
@@ -25,7 +25,8 @@ func registerGRPCServices(r *mux.Router, server *grpc.Server) {
 		runtime.WithIncomingHeaderMatcher(CustomMatcher),
 	)
 
-	_ = grpcGatewayHandler.HandlePath("POST", `/v1/file/upload/{allocation}`, uploadHandler)
+	_ = grpcGatewayHandler.HandlePath("POST", `/v1/file/upload/{allocation}`,
+		upload(common.UserRateLimit(common.ToJSONResponse(WithConnection(UploadHandler)))))
 
 	blobbergrpc.RegisterBlobberServiceServer(server, blobberService)
 	_ = blobbergrpc.RegisterBlobberServiceHandlerServer(context.Background(), grpcGatewayHandler, blobberService)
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -125,6 +125,18 @@ func setupHandlerContext(ctx context.Context, r *http.Request) context.Context {
 	return ctx
 }
 
+func setupGRPCMuxHandlerContext(ctx context.Context, r *http.Request, pathParams map[string]string) context.Context {
```
kushthedude, line 128: Why? We already have a similar function, look at the functions at `grpc_handler.go`

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
```diff
@@ -83,3 +83,10 @@ func CustomMatcher(key string) (string, bool) {
 		return runtime.DefaultHeaderMatcher(key)
 	}
 }
+
+func uploadHandler(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
```
kushthedude, line 87: No need for this actually

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
```diff
@@ -18,14 +18,14 @@ type gRPCHeaderMetadata struct {
...
 func registerGRPCServices(r *mux.Router, server *grpc.Server) {
 	blobberService := newGRPCBlobberService()
 	r.Use(Middleware2("ds"))
 	grpcGatewayHandler := runtime.NewServeMux(
-								runtime.WithIncomingHeaderMatcher(CustomMatcher),
-							)
+		runtime.WithIncomingHeaderMatcher(CustomMatcher),
+	)
 
+	_ = grpcGatewayHandler.HandlePath("POST", `/v1/file/upload/{allocation}`, uploadHandler)
```
kushthedude, line 28: Can we not re-use the already implemented HTTP handler in the `handler.go` file?

kushthedude (changes requested):

code/go/0chain.net/blobbercore/handler/handler.go
senpathi: Yes, this to get path params or we need to change that function to get map[string]string

kushthedude: We already have function to do this see this line in the `grpc_handlers.go` `setupGrpcHandlerContext(ctx, getGRPCMetaDataFromCtx(ctx))`

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
senpathi: where should this handler be implemented?

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
senpathi: Yes, we can rewrite StorageHandler Methods for it

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
kushthedude: No actually we already have this implemented, we can just use that. See the `UploadHandler` in the handler.go file.

kushthedude (approved):

PR #325, opened 2021-08-28
Fix json tag issue

lpoli:
The json tag that had been used broke old api that used default value(struct field).
To fix this issue New struct(as Gorm smart select field) is created leaving Ref struct as it is.


cnlangzi:
@lpoli  and me had some discussion about this feature. I will refactor some code instead of code review first.

cnlangzi (approved):
let's merged it to fix tag issues first. we will raise new issue to discuss how to implement api with new code style.

PR #326, opened 2021-08-28
multiple handler funcs for  upload handler with different http methods

senpathi:
Multiple method support from grpc-gateway handlers to upload handler

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/grpc_helpers.go
+++ b/code/go/0chain.net/blobbercore/handler/grpc_helpers.go
@@ -36,7 +36,19 @@ func registerGRPCServices(r *mux.Router, server *grpc.Server) {
 			common.UserRateLimit(common.ToJSONResponse(WithConnection(UploadHandler)))(w, r)
 		})
 
-	_ = grpcGatewayHandler.HandlePath("POST", "/v1/file/download/{allocation}",
+	_ = grpcGatewayHandler.HandlePath("PUT", "/v1/file/upload/{allocation}",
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/blobbercore/handler/grpc_helpers.go
```diff
@@ -36,7 +36,19 @@ func registerGRPCServices(r *mux.Router, server *grpc.Server) {
 			common.UserRateLimit(common.ToJSONResponse(WithConnection(UploadHandler)))(w, r)
 		})
 
-	_ = grpcGatewayHandler.HandlePath("POST", "/v1/file/download/{allocation}",
+	_ = grpcGatewayHandler.HandlePath("PUT", "/v1/file/upload/{allocation}",
```
kushthedude, line 39: Why the error is ignored in each place?
```suggestion
	err = grpcGatewayHandler.HandlePath("PUT", "/v1/file/upload/{allocation}",
		if err != nil {
		log.Errorf("failed to add custom route: %v", err)
		panic(err)
	}
```

kushthedude (changes requested):
This piece of code might have been missed from my previous reviews, would you mind adding error checks in all the three places where we have been injecting custom routes?

code/go/0chain.net/blobbercore/handler/grpc_helpers.go
senpathi: okay sure

kushthedude (approved):

PR #332, opened 2021-09-03
Return err

niki290595:


code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -5,20 +5,20 @@ import (
...
-	blobbergrpc "github.com/0chain/blobber/code/go/0chain.net/blobbercore/blobbergrpc/proto"
-	"github.com/pkg/errors"
 	"math"
+	"net/http"
+	"path/filepath"
+	"strconv"
 	"strings"
 
+	blobbergrpc "github.com/0chain/blobber/code/go/0chain.net/blobbercore/blobbergrpc/proto"
+	"github.com/pkg/errors"
```
cnlangzi, line 15: it seems is replaced with github.com/0chain/errors. please double check it

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -327,6 +327,7 @@ func (fsh *StorageHandler) AddCollaborator(ctx context.Context, request *blobber
 
 func (fsh *StorageHandler) GetFileStats(ctx context.Context, request *blobbergrpc.GetFileStatsRequest) (interface{}, error) {
 	// todo(kushthedude): generalise the allocation_context in the grpc metadata
+	Logger.Info("verifyAllocation...")
```
cnlangzi, line 330: I think it is debugging info on local.  It is better to remove it. Because we can't get any useful information from a grafana with `verfiyAllocation....`

cnlangzi (changes requested):
could you remove those logger.Info without any detail information. eg

```
Logger.Info("GetListingData...")

	Logger.Info("GetFileStats...")
```

They aren't helpful for debugging on server, and will slow api response.

PR #339, opened 2021-09-08
Fix file stats

niki290595:


cnlangzi (changes requested):
why ci check is not passed. please check it. thanks

PR #348, opened 2021-09-23
Fix/clean GitHub action

cnlangzi:
- enabled `Unit tests`
- clean script


service-0chain:
@cnlangzi 
Use self-hosted runners instead of github runners.

PR #353, opened 2021-09-23
Grpc v2 fix trans bug

niki290595:


code/go/0chain.net/core/chain/entity.go
```diff
@@ -21,8 +18,6 @@ func SetServerChain(c *Chain) {
 
 /*GetServerChain - returns the chain object for the server chain */
 func GetServerChain() *Chain {
-	logging.Logger.Info("call chain.GetServerChain",
```
kushthedude, line 24: Why did you remove this log?

niki290595: it clogs the log

kushthedude (approved):

PR #373, opened 2021-09-29
Respond with certain level if level is present

lpoli:
This change modifies use of `level` field of `reference_objects` table. It is used to query refs that are of certain level.


cnlangzi (approved):
LGTM

PR #380, opened 2021-10-05
Migrate marketplace endpoint to grpc

amanjain97:
## WHAT
1. Migrate `/v1/marketplace/shareinfo/{allocation}` HTTP endpoint to GRPC endpoint. 
Related PR: https://github.com/0chain/blobber/pull/311

2. Reformat code 
3. Update buf.yaml to ignore field deletion with reserved keyword.


code/go/0chain.net/blobbercore/blobbergrpc/proto/blobber_contract.proto
```diff
@@ -370,6 +370,8 @@ message MarketplaceShareInfoRequest {
 }
 
 message MarketplaceShareInfoResponse {
-  string response_status = 1;
+  reserved 1;
```
kushthedude, line 373: You can remove the reserved here as the following proto is not in prod, I dont think we may get any backward compatibility issue

kushthedude (approved):

PR #386, opened 2021-10-12
Expanding unit-tests && fixing issue with renaming dirs

### LINKED ISSUE #382: RenameDir is not working
There are few methods like move or copy, but non of them working for Dirs. I am assigning this ticket to myself as already started with it ad it's blocking mobile team from release candidate

moldis:
This branch is under WIP yet
Related to https://github.com/0chain/blobber/issues/382


code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -198,6 +198,9 @@ func GetSubDirsFromPath(p string) []string {
 	parent = filepath.Clean(parent)
 	subDirs := make([]string, 0)
 	for len(cur) > 0 {
+		if cur == "." {
```
lpoli, line 201: Don't understand what kind of issue it fixes. Can you please show when will `cur == "."` be useful?

Also this function is very inefficient. I know its written by Sarath Ambati.
It should be splitted by `strings` package. `strings.Split(s, "/")`

code/go/0chain.net/blobbercore/allocation/renamefilechange.go
```diff
@@ -68,6 +68,12 @@ func (rf *RenameFileChange) ProcessChange(ctx context.Context, change *Allocatio
 			return nil, common.NewError("invalid_reference_path", "Invalid reference path from the blobber")
 		}
 	}
+
+	if len(dirRef.Children) == 0 {
```
lpoli, line 72: If `dirRef` have no children then I doubt code will reach this line.
It should return error when calling `GetObjectTree` in line 28

Also this function can be highly optimized. 
The `for` loop in line 54 have a single purpose of finding `ref` in rootRef so some other child `refs` can be removed and then calculate Hash of all affected directory `ref`. This `for` loop is very inefficient. We need to refactor it.
Please verify.

lpoli (changes requested):
One function that needs optimization is `GetReferencePath`. What should be done is, refactor `GetReferencePath` and make respective changes to all the caller.

Anyway `for` loop in line 56 is very inefficient. It should be refactored.

cnlangzi (approved):

PR #387, opened 2021-10-13
fix(copy):0chain/zboxcli:115 fixed hash issue on copy

cnlangzi:
fixed hash bug for https://github.com/0chain/zboxcli/issues/115


code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -259,7 +259,7 @@ func (fr *Ref) GetFileHashData() string {
 	if len(fr.Attributes) == 0 {
 		fr.Attributes = datatypes.JSON("{}")
 	}
-	hashArray := make([]string, 0)
+	hashArray := make([]string, 0, 11)
```
lpoli, line 262: Why 11? Its not required. Even with capacity of 11 it's not enough for allocation id, type, name , etc.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -271,17 +271,17 @@ func (fr *Ref) GetFileHashData() string {
 	hashArray = append(hashArray, fr.ActualFileHash)
 	hashArray = append(hashArray, string(fr.Attributes))
 	hashArray = append(hashArray, strconv.FormatInt(fr.ChunkSize, 10))
-	return strings.Join(hashArray, ":")
+
+	hashData := strings.Join(hashArray, ":")
```
lpoli: IMO initializing new variable just to return it is redundant.

code/go/0chain.net/blobbercore/allocation/copyfilechange.go
```diff
@@ -120,6 +120,7 @@ func (rf *CopyFileChange) processCopyRefs(ctx context.Context, affectedRef *refe
 		newFile.ActualThumbnailSize = affectedRef.ActualThumbnailSize
 		newFile.EncryptedKey = affectedRef.EncryptedKey
 		newFile.Attributes = datatypes.JSON(string(affectedRef.Attributes))
+		newFile.ChunkSize = affectedRef.ChunkSize
```
lpoli, line 123: It seems only this change was required. I wonder how it was working previously.

cnlangzi: It didn't work after ChunkedUpload is merged.

code/go/0chain.net/blobbercore/reference/ref.go
cnlangzi: there are only 11 items in this array.

lpoli:
There was missing component(ChunkSize) for Hash Calculation which resulted in allocation mismatch error. It is fixed by simply adding that component.

PR #390, opened 2021-10-14
:bug: Reject rename if file already exists

### LINKED ISSUE #389: rename should be rejected if target filename exists
it will be overwritten if target filename exists now.

moldis:
Related to https://github.com/0chain/blobber/issues/389


code/go/0chain.net/blobbercore/allocation/renamefilechange.go
```diff
@@ -29,6 +29,12 @@ func (rf *RenameFileChange) ProcessChange(ctx context.Context, change *Allocatio
 	if err != nil {
 		return nil, err
 	}
+
+	_, err = reference.GetObjectTree(ctx, rf.AllocationID, rf.NewName)
```
cnlangzi: It is too heavy to load so many nodes for it?

```
	if path != "/" {
		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "/%"), allocationID)
	} else {
		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "%"), allocationID)
	}
```

do you think it is better to add a simple check function for better sql performance?

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/allocation/renamefilechange.go
moldis: @cnlangzi Yes, the whole function is very messy and hard to read. I created separate issue to refactor it

code/go/0chain.net/blobbercore/allocation/renamefilechange.go
```diff
@@ -25,10 +25,16 @@ func (rf *RenameFileChange) DeleteTempFile() error {
 }
 
 func (rf *RenameFileChange) ProcessChange(ctx context.Context, change *AllocationChange, allocationRoot string) (*reference.Ref, error) {
+	exists, _ := reference.PathExists(ctx, rf.AllocationID, rf.NewName)
+	if exists {
+		return nil, common.NewError("invalid_reference_path", "Invalid new name of path, it's already exists")
+	}
```
kushthedude:
```suggestion
	isFilePresent, _ := reference.PathExists(ctx, rf.AllocationID, rf.NewName)
	if isFilePresent {
		return nil, common.NewError("invalid_reference_path", "file already exists")
	}
```

kushthedude (dismissed):
Looks good to me @moldis, have made a small grammatical correction. Btw do you mind adding a small unit test to verify this fix?

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -78,6 +78,30 @@ func GetReferencePathFromPaths(ctx context.Context, allocationID string, paths [
...
+
+	if path == "." || path == "/" {
+		return false, nil
+	}
+
+	var refs []Ref
+	db := datastore.GetStore().GetTransaction(ctx)
+
+	ref := Ref{Path: path, AllocationID: allocationID}
+	db = db.Where(ref).Or("path LIKE ? AND allocation_id = ?", (path + "/%"), allocationID)
```
cnlangzi: do you think `db = db.Where(ref)` is enough?

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -78,6 +78,30 @@ func GetReferencePathFromPaths(ctx context.Context, allocationID string, paths [
...
+
+	err := db.Order("level, lookup_hash").Preload(ref.TableName(), func(tx *gorm.DB) *gorm.DB {
```
cnlangzi: `order by` and `preload` should be unnecessary here? please double check it. thanks.

cnlangzi (changes requested):

cnlangzi (approved):

PR #394, opened 2021-10-19
Signature verification in file meta api to ensure authorized access to allocation owner or a registered collaborator

### LINKED ISSUE #239: Not all handler function verifies client signature
ObjectTree, ObjectPath, etc. have all implemented signature verification but following handler does not require client signature.
1. _/v1/file/meta/{allocation}_
2. _/v1/file/list/{allocation}_

There maybe other requests handler that does not require client signature.

### LINKED ISSUE #327: Security: Some file metadata endpoints can be queried without a signature
Similar to https://github.com/0chain/blobber/issues/48
The following endpoints can be queried with no authentication:
/v1/file/meta
/v1/file/list
Potentially leaking file metadata to anyone with the allocation id 

Solution:
Add X-App-Client-Signature header to the endpoints above as in https://github.com/0chain/blobber/issues/48
The signature will be generated on the client by BLS signing the hash of the allocation id
Possible clients that will need updated 
-Go SDK
-JS SDK
-0CLI
-Block-explorer

iamrz1:


code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -74,6 +75,29 @@ func SetupHandlers(r *mux.Router) {
...
+		if allocationObj.Expiration < common.Now() {
+			return nil, common.NewError("verify_allocation", "use of expired allocation")
+		}
+
+		valid, err := verifySignatureFromRequest(allocationString, r.Header.Get(common.ClientSignatureHeader), allocationObj.OwnerPublicKey)
+		if !valid || err != nil {
+			return nil, common.NewError("invalid_signature", "Invalid signature")
+		}
+
+		return handler(ctx, r)
```
cnlangzi: please save `allocationObj` in context, and in handler , we can get it from context directly. it will save some time to load it again. Because almost handlers will read it again.

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/handler/handler.go
iamrz1: Yes it will also improve the dry-ness of the code. In that case we'll have to remove all the allocation initialization code from the handlers. That'd be a lot of code rewrites.

cnlangzi: @iamrz1  migrate them one by one. It is unnecessary to get it done in one PR if it is too much. But it should be fixed on handlers you are woking on now.

iamrz1: I have moved allocation object to context for file-meta/ file-stats handlers. take a look and let me know what you think.

code/go/0chain.net/blobbercore/handler/grpc_handler.go
```diff
@@ -45,7 +45,7 @@ func (b *blobberGRPCService) GetFileMetaData(ctx context.Context, req *blobbergr
 		"auth_token": {req.AuthToken},
 	}
 
-	resp, err := FileMetaHandler(ctx, r)
+	resp, err := WithReadOnlySignatureVerification(FileMetaHandler)(ctx, r)
```
iamrz1: @kushthedude what do you think of this approach to address http middlewares in grpc handlers?

cnlangzi: @kushthedude  do you mean move all middleware to grpc handler instead of http hanlder?

code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -115,6 +116,79 @@ func WithConnection(handler common.JSONResponderF) common.JSONResponderF {
...
+	}
+}
+
+func WithReadOnlySignatureVerification(handler common.JSONResponderF) common.JSONResponderF {
+	return func(ctx context.Context, r *http.Request) (interface{}, error) {
+		var err error
+		ctx = setupHandlerContext(ctx, r)
```
cnlangzi: you can merge them into one method without another boxing/unboxing from context.

cnlangzi: `setupHandlerContext` and `performSignatureVerification`

cnlangzi: for example: https://github.com/0chain/blobber/blob/c82cb186436acf6d78bb611d3add8c685978f47d/code/go/0chain.net/blobbercore/handler/handler.go#L76

cnlangzi: what difference between `WithReadWriteSignatureVerification` and `WithReadOnlySignatureVerification`?

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/handler/handler.go
iamrz1: `setupHandlerContext` is being used in other handler functions right now. so it has to stay that way. I can merge `performSignatureVerification` in the parent function, but I kept it that way to keep it available for non-readonly operations using `WithReadWriteSignatureVerification` middleware.

`WithReadOnlySignatureVerification` checks if client has read permission, `WithReadWriteSignatureVerification` checks (according to my interpretation) if client has write permission too. We might use this middleware when we replace existing repetition in handlers with middlewares.

code/go/0chain.net/blobbercore/handler/grpc_handler.go
iamrz1: No, since I'm doing signature verification from http middlewares, grpc handlers also need to call the http middleware. This is my solution to that problem.

code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -115,6 +116,62 @@ func WithConnection(handler common.JSONResponderF) common.JSONResponderF {
...
+
+		allocationString := allocationInContext.(string)
+
+		allocationObject, err := verifyAllocation(ctx, allocationString, false)
+		if err != nil {
+			return nil, common.NewErrorf("verify_allocation", "verifying allocation transaction error: %v", err)
+		}
+
+		signatureHeader := ctx.Value(constants.ContextKeyClientSignatureHeaderKey)
+		if signatureHeader == nil {
```
kushthedude: Please lets not add additional wrapper over the handlers, gRPC code already takes care of this so lets not make more logical changes here & simply do the signature verification in the handler logic

code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -115,6 +116,62 @@ func WithConnection(handler common.JSONResponderF) common.JSONResponderF {
...
+		var vars = mux.Vars(r)
+		ctx = context.WithValue(ctx, constants.ContextKeyClient, r.Header.Get(common.ClientHeader))
+		ctx = context.WithValue(ctx, constants.ContextKeyClientKey, r.Header.Get(common.ClientKeyHeader))
+		ctx = context.WithValue(ctx, constants.ContextKeyAllocation, vars["allocation"])
+		// signature is not requered for all requests, but if header is empty it won`t affect anything
+		ctx = context.WithValue(ctx, constants.ContextKeyClientSignatureHeaderKey, r.Header.Get(common.ClientSignatureHeader))
+
+		allocationInContext := ctx.Value(constants.ContextKeyAllocation)
+		if allocationInContext == nil {
+			return nil, common.NewErrorf("invalid_allocation", "Missing allocation in context")
```
kushthedude: This is unrelated to signature verification, either lets do it in other PR or drop it as is as the gRPC takes in allocation & all the other headers into the context & pass through the values using context https://github.com/0chain/blobber/blob/fd2592402e5c95d47f80ea4c3710730d0509c543/code/go/0chain.net/blobbercore/handler/grpc_helpers.go#L32

cnlangzi (dismissed):
LGTM. please fix go.mod/go.sum.

code/go/0chain.net/blobbercore/handler/grpc_handler.go
kushthedude: Why would grpc call http?

iamrz1: I'm not sure what you mean by this but `resp, err := FileMetaHandler(ctx, r)` is a call to a http handler named `FileMetaHandler`, right? Ergo, my interpretation is that grpc handler is making calls to http habdler.

I just added the same middleware that http handler is using now.

code/go/0chain.net/blobbercore/handler/handler.go
iamrz1: `getGRPCMetaDataFromCtx` and the code block in this middleware does not perform the same task, take a look at the codebase, and you'll see.

And codes that are seemingly unrelated to signature verification (getting and setting values in ctx) are part of planned streamlining handler codebase (ie. avoid getting allocation object multiple times or setting values in ctx in different places) as almost all the handlers are performing signature verification anyway.

code/go/0chain.net/blobbercore/handler/handler.go
iamrz1: There's nothing here that repeats codes from gRPC. And there's no logical change, only the verification part is being done before executing the handler.
The reason why the signature verification code is being moved to a middleware is, verification check is a crucial security issue and there has been a separate PR before to address endpoints that were missing it. But the file-stats and file-meta apis still missed verification check because of oversight. Moving verification to a middleware will make it glaringly obvious which endpoints are missing this security check.

code/go/0chain.net/blobbercore/handler/handler.go
kushthedude: getGRPCMetaDataFromCtx, this takes the value of the allocation from the request, please read the complete code block.

> (ie. avoid getting allocation object multiple times or setting values in ctx in different places) 

The above function does it correctly, anyways go ahead as you like as ultimately grpc is going to replace all this cluttered middlewares

kushthedude (dismissed):
Lets merge this as it is a crucial security fix & regarding the discussion on middleware gRPC will have to take rid of them and abstract them into helper utilities. The only reason I was not in favour of middleware because if there was a helper we could have reused it in grpc preprocessing but for middlewares we need to write a custom helper to replace it 😓 


@cnlangzi please do approve and merge

iamrz1:
It appears that the issue with go-sdk still has not been resolved. Blobber is still failing tests.

cnlangzi:
critical security issue, get it merged first.

PR #395, opened 2021-10-19
re-implemented `createdir` and added development toolkits

### LINKED ISSUE #230: [FEATURE] use database migrate tool instead of pure sql scripts
it is better to migrate database with migrate tool than docker `postgres-post`.

- https://flywaydb.org/
- https://github.com/golang-migrate/migrate
- https://gorm.io/docs/migration.html
- https://github.com/go-gormigrate/gormigrate

And db should be migrated automatically on blobber startup.

### LINKED ISSUE #329: createDir: Incorrect values are saved in reference_objects and break challenge
https://github.com/0chain/blobber/blob/7370c1e346733c3a31285bd0e75be1a87796c2ab/code/go/0chain.net/blobbercore/allocation/newfilechange.go#L89

It should be 

```
newDir.ActualFileSize = 0
newDir.Size = 0
newDir.NumBlocks = 0
````

Because it is new empty directory.

### PR DESCRIPTION
**Pull request by cnlangzi**
- Added AutoMigrate feature. see  [detail](https://github.com/0chain/blobber/blob/0aea42103f924138d3cf1896569e6b44e85e8e45/code/go/0chain.net/blobbercore/models)
- Added new blobber hanlder development toolkit for better readable and performance [detail](https://github.com/0chain/blobber/blob/0aea42103f924138d3cf1896569e6b44e85e8e45/code/go/0chain.net/blobbercore/handler/handler.go#L76)
   + Removed transaction from request `WithConnection`.
   + Removed transaction from request `WithReadonlyConnection` 
   + Added [DryRun](https://github.com/0chain/blobber/blob/0aea42103f924138d3cf1896569e6b44e85e8e45/code/go/0chain.net/blobbercore/allocation/dao.go#L42) to prepare statement for speed up sql call
   + Added new [DAO](https://github.com/0chain/blobber/blob/0aea42103f924138d3cf1896569e6b44e85e8e45/code/go/0chain.net/blobbercore/allocation/dao.go#L14) to disable ORM feature.  
   + Add `WithAuth` to verify allocation and signature , and pass it directly without boxing/unboxing from context.Context
- Reimplemented `CreateDir`, and get `zbox list` worked with it.

### CODE REVIEW

File: code/go/0chain.net/blobbercore/handler/handler.go [hunk 1]
```diff
@@ -72,6 +69,12 @@ func SetupHandlers(r *mux.Router) {
 
 	//marketplace related
 	r.HandleFunc("/v1/marketplace/shareinfo/{allocation}", common.ToJSONResponse(WithConnection(MarketPlaceShareInfoHandler)))
+
+	//New API handlers for better readable and performance
+
+	//directory
+	r.HandleFunc("/v1/dir/{allocation}", WithJSON(CreateDir)).Methods(http.MethodPut)
```
moldis, line 76: @cnlangzi PUT methods is for updates mostly, this one is create should be POST

code/go/0chain.net/blobbercore/reference/ref_walker.go
```diff
@@ -0,0 +1,146 @@
...
+
+type RefWalker struct {
+	items  []string
+	length int
+	index  int
+}
+
+// Name get current dir name
+func (d *RefWalker) Name() string {
+	if d == nil {
```
moldis, line 30: Not sure in what case d can be nil?

code/go/0chain.net/blobbercore/handler/handler.go
cnlangzi: @moldis  It is designed to work for CREATE/UPDATE. please see directory_test.go

code/go/0chain.net/blobbercore/reference/ref_walker.go
cnlangzi: It prevents crash if some guy write bad code.

code/go/0chain.net/blobbercore/models/migration.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/0chain/blobber/code/go/0chain.net/core/logging"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+const (
+	TableNameMigration = "migrations"
+)
+
+type Migration struct {
```
moldis, line 17: I thought gorm supports auto migration with models, what is the purpose of this file?

code/go/0chain.net/blobbercore/models/reference_object.go
```diff
@@ -0,0 +1,50 @@
...
+	ActualThumbnailSize int64          `gorm:"column:actual_thumbnail_size"`
+	ActualThumbnailHash string         `gorm:"column:actual_thumbnail_hash"`
+	EncryptedKey        string         `gorm:"column:encrypted_key"`
+	Attributes          datatypes.JSON `gorm:"column:attributes"`
+
+	OnCloud bool `gorm:"column:on_cloud"`
+
+	CreatedAt time.Time      `gorm:"column:created_at"`
+	UpdatedAt time.Time      `gorm:"column:updated_at"`
+	DeletedAt gorm.DeletedAt `gorm:"column:deleted_at"` // soft deletion
```
moldis: Instead of defining these fields for each type, you can define type once `type ModelWithTS struct {
	CreatedAt time.Time `gorm:"column:created_at;autoCreateTime"`
	UpdatedAt time.Time `gorm:"column:updated_at;autoUpdateTime:milli"`
}`

And then add datastore.ModelWithTS (example) for each type you want to include CreateAt/UpdateAt and not sure what's the DeleteAt ?

code/go/0chain.net/blobbercore/models/migration.go
cnlangzi: I have tested it .It doesn't work properly for index and constraints. And it is stupid on column type checking. so have to write it by own function.

code/go/0chain.net/blobbercore/reference/dao.go
```diff
@@ -0,0 +1,80 @@
...
+	it := &models.ReferenceObject{}
+
+	result := db.Table(models.TableNameReferenceObject).Where(SQLWhereGetByAllocationTxAndPath, allocationTx, path).First(it)
+
+	if result.Error == nil {
+		return it, nil
+	}
+
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+		return it, errors.Throw(constants.ErrEntityNotFound, "alloction_id: "+allocationTx+" path: "+path)
```
moldis: typo here: alloction_id

code/go/0chain.net/blobbercore/reference/ref_walker.go
moldis: @cnlangzi Still not clear, if RefWalker is nil - how anyone can call Name() or anything else?

code/go/0chain.net/blobbercore/models/reference_object.go
cnlangzi: fixed it. deleted_at already is defined in db. I have no chance to check how it works. It is only a mapping between db and model.

code/go/0chain.net/blobbercore/models/migration.go
moldis: @cnlangzi It seems working fine on the 0box server (it's been many migrations so far), may I know what issues you faced with testing it in index and constraints?

moldis (approved):
Looks good

code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -22,7 +22,9 @@ func (store *postgresStore) Open() error {
 		"host=%v port=%v user=%v dbname=%v password=%v sslmode=disable",
 		config.Configuration.DBHost, config.Configuration.DBPort,
 		config.Configuration.DBUserName, config.Configuration.DBName,
-		config.Configuration.DBPassword)), &gorm.Config{})
+		config.Configuration.DBPassword)), &gorm.Config{
+		SkipDefaultTransaction: true, // https://gorm.io/docs/performance.html#Disable-Default-Transaction
```
lpoli, line 26: We might desire data consistency to performance. So I don't think we should skip default transaction.  Gorm document clearly says that if we don't require data consistency then we can increase performance by skipping default transaction.

code/go/0chain.net/blobbercore/filestore/fs_store.go
```diff
@@ -450,14 +450,6 @@ func (fs *FileFSStore) DeleteFile(allocationID string, contentHash string) error
 	return os.Remove(fileObjectPath)
 }
 
-func (fs *FileFSStore) CreateDir(dirName string) error {
```
lpoli, line 453: Deleting this function under different commit. Commit was related to automigration but this function deletion makes less sense under this commit.

go.mod
```diff
@@ -37,4 +37,4 @@ require (
 	nhooyr.io/websocket v1.8.7 // indirect
 )
 
-// replace github.com/0chain/gosdk => ../gosdk
+replace github.com/0chain/gosdk => ../gosdk
```
lpoli: Just a reminder to comment out this line

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: - In old code, a db transaction is started at every request begin. 
- In new code, we use db.Transaction(func(tx *gorm.DB) error) to start transaction. 
so it is unnecessary to let gorm to do it automatically. Actually, transaction is unnecessary for performance if only a record is added/updated/deleted .

code/go/0chain.net/blobbercore/filestore/fs_store.go
cnlangzi: commit is unnecessary for `CreateDir`.  and `CreateDir` is unnecessary to do anything on local disk. so we don't need it in our filestore interface.

go.mod
cnlangzi: will fix it before merging

code/go/0chain.net/blobbercore/models/reference_object.go
lpoli: Gorm by default implements soft delete, that means, it will update deleted_at column and will only show rows which have deleted_at column value to be null.
We can use other method to retrieve deleted rows. I think its `unscoped` method.

code/go/0chain.net/blobbercore/models/migration.go
```diff
@@ -0,0 +1,97 @@
...
+	CreatedAt time.Time `gorm:"column:created_at"`
+}
+
+func (m *Migration) After(latest *Migration) bool {
+	currentVersions := strings.Split(m.Version, ".")
+	latestVersions := strings.Split(latest.Version, ".")
+
+	for i := 0; i < 3; i++ {
+		if currentVersions[i] > latestVersions[i] {
+			return true
```
lpoli, line 33: It seems version number priority is not considered. Comparing 0.2.3 with 1.1.3 it will return true.

code/go/0chain.net/blobbercore/models/migration.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+}
+
+var releases = []Migration{
+	Migration{
+		Version:   "0.1.0",
+		CreatedAt: time.Date(2021, 10, 15, 0, 0, 0, 0, time.UTC),
+	},
+}
```
lpoli, line 109: If we have to add new scripts to update table schema then it is no different than SQL code. If we cannot make proper use of gorm fields then we should use traditional way(with .sql file)

code/go/0chain.net/blobbercore/allocation/dao.go
```diff
@@ -0,0 +1,52 @@
...
+const (
+	SQLWhereGetByTx = "allocations.tx = ?"
+)
+
+// DryRun  Creates a prepared statement when executing any SQL and caches them to speed up future calls
+// https://gorm.io/docs/performance.html#Caches-Prepared-Statement
+func DryRun(db *gorm.DB) {
+
+	// Session mode
+	tx := db.Session(&gorm.Session{PrepareStmt: true})
```
lpoli: Does it cache query after session is complete?
I see that you have `PrepareStmt` set to true while opening postgres db which is global .

code/go/0chain.net/blobbercore/allocation/zcn.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/0chain/blobber/code/go/0chain.net/core/chain"
+	"github.com/0chain/blobber/code/go/0chain.net/core/node"
+	"github.com/0chain/blobber/code/go/0chain.net/core/transaction"
+	"github.com/0chain/errors"
+	"github.com/0chain/gosdk/constants"
+	"gorm.io/gorm"
+)
+
+// SyncAllocation try to pull allocation from blockchain, and insert it in db.
+func SyncAllocation(allocationTx string) (*models.Allocation, error) {
```
lpoli, line 17: This refactorization has made code more readable but performance wise it has no effect. Also it is under auto-migration commit.

code/go/0chain.net/blobbercore/handler/handler.go
lpoli: I think for create and update both we use post rather than put

code/go/0chain.net/blobbercore/datastore/postgres.go
lpoli: If we are using db.Transaction the we are using transaction already. It only makes sense to use SkipDefaultTransaction if we don't use transaction for read operation.

code/go/0chain.net/blobbercore/filestore/fs_store.go
lpoli: My point is that it will be difficult to review if you put irrelevant file changes under this commit.

This commit message is about automigration but deleting createDir function has no purpose in it.

lpoli (changes requested):
I suppose by auto-migration we mean adding/updating table schema, adding/updating indexes, etc. Also the migration code has not taken data migration into consideration so this is related to table schema only.

If that can be achieved by gorm fields then it is fine but if we need to create newer scripts then I don't think it is easier approach.

Also please add appropriate files changes under some commit. This commit is related to auto-migration of postgresql schema but has other changes as well.

code/go/0chain.net/blobbercore/handler/handler.go
cnlangzi: `The difference between PUT and POST is that PUT is idempotent: calling it once or several times successively has the same effect (that is no side effect), whereas successive identical POST requests may have additional effects, akin to placing an order several times`  https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT#:~:text=The%20difference%20between%20PUT%20and,placing%20an%20order%20several%20times.

that is why I think it is better to implemented as `PUT` instead of `POST`. Because it is a idempotent operation here.

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: NO. please check it on. https://gorm.io/docs/performance.html#Disable-Default-Transaction

`GORM perform write (create/update/delete) operations run inside a transaction to ensure data consistency, which is bad for performance, you can disable it during initialization`

it is not designed for `read`.

code/go/0chain.net/blobbercore/models/migration.go
cnlangzi: Yes. because index is changed.so it must be migrated. 
but 0.2.3 is incorrect one. version is set by developer. why he set it as 0.2.3 if we have 1.1.3 now? 

Anyway, it will work properly if it only has migration script for index.

code/go/0chain.net/blobbercore/filestore/fs_store.go
cnlangzi: I am try to re-implement CreateDir, but it asks to migrate schema to add index. that is why `AutoMigrate` is added here.

code/go/0chain.net/blobbercore/models/migration.go
cnlangzi: please have a look at [README.md](https://github.com/0chain/blobber/blob/fix/break_challenge_on_creatdir/code/go/0chain.net/blobbercore/models/README.md#migrate-tablecolumn-in-gormautomigrate)

cnlangzi: gorm AutoMigrate is stupid with some cases.
- reference.PathLevel, it can't know the type as well. so migrate script always be executed. the problem is the migrate script doesn't work. so it always is migrated again and again......
- add foreign key , you have to add embedded objects to let gorm know how to migrate it. eg.   allocations.terms. But advance ORM feature is removed from blobber code.  It is better to let ORM worked with plain object for better performance

so gorm AutoMigrate is not good for everything. We just do something manually if it doesn't work properly with gorm.

code/go/0chain.net/blobbercore/allocation/zcn.go
cnlangzi: have a look at `VerifyAllocationTransaction`. 
- for new allocation, it query db twice with transaction.
- for old allocation , it load allocation and terms with transaction too. so 2 db operations too. but terms are unnecessary for almost request

In new code, Only one query operation without starting a heavy transaction for new/old allocation. and terms will not be loaded too . 
It will improve a lot of performance on it.  Because almost of command will ask to load allocation. if it saves 100ms, 10 times mean 1s.  I don't think it has no effect on performance.  @guruhubb asks to develop blobber as high concurrent request system. Roman is not built one day. We need improve performance as much as possible.

cnlangzi:
> I suppose by auto-migration we mean adding/updating table schema, adding/updating indexes, etc. Also the migration code has not taken data migration into consideration so this is related to table schema only.
> 
> If that can be achieved by gorm fields then it is fine but if we need to create newer scripts then I don't think it is easier approach.
> 
> Also please add appropriate files changes under some commit. This commit is related to auto-migration of postgresql schema but has other changes as well.


This PR is submitted for `CreateDir`. Other changes are developed for it too. if we split them into other PRs again, our code will still keep smell. 
I would suggest to improve it pice by pice when we are fixing bugs or adding feature if it is possible. 

It is a big challenge if someone assign me an issue to refactor all db code at the same time.  It is real difficult to get it done in time. And we have to stop to work on any features /bugs .

code/go/0chain.net/blobbercore/allocation/dao.go
cnlangzi: good question. I need to go through gorm code to double check it.  DryRun is designed for speed up first db call in request.

cnlangzi: It seems is designed as global objects.so it should work with session and global mode. but let's change it with global mode. it should be better than session mode.

https://github.com/go-gorm/gorm/blob/38e55f1117858e49c372e0e7a9f22f9761c7cd8b/gorm.go#L176
```
preparedStmt := &PreparedStmtDB{
		ConnPool:    db.ConnPool,
		Stmts:       map[string]Stmt{},
		Mux:         &sync.RWMutex{},
		PreparedSQL: make([]string, 0, 100),
	}
	db.cacheStore.Store(preparedStmtDBKey, preparedStmt)

	if config.PrepareStmt {
		db.ConnPool = preparedStmt
	}
```

https://github.com/go-gorm/gorm/blob/38e55f1117858e49c372e0e7a9f22f9761c7cd8b/gorm.go#L244
```
if config.PrepareStmt {
		if v, ok := db.cacheStore.Load(preparedStmtDBKey); ok {
			preparedStmt := v.(*PreparedStmtDB)
			tx.Statement.ConnPool = &PreparedStmtDB{
				ConnPool: db.Config.ConnPool,
				Mux:      preparedStmt.Mux,
				Stmts:    preparedStmt.Stmts,
			}
			txConfig.ConnPool = tx.Statement.ConnPool
			txConfig.PrepareStmt = true
		}
	}
```

code/go/0chain.net/validatorcore/storage/models.go
```diff
@@ -69,9 +69,12 @@ func (r *DirMetaData) GetHash() string {
 
 func (r *DirMetaData) CalculateHash() string {
 
-	childHashes := make([]string, len(r.Children))
-	for index, childRef := range r.Children {
-		childHashes[index] = childRef.GetHash()
+	childHashes := make([]string, 0, len(r.Children))
+	for _, childRef := range r.Children {
+		if len(childRef.GetHash()) > 0 {
```
lpoli, line 74: When will it be equal to 0?

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -23,31 +26,46 @@ func GetReferencePath(ctx context.Context, allocationID string, path string) (*R
...
+
+			if len(p) == 0 {
+				break
+			}
+
+			if ok := pathsAdded[p]; !ok {
+
+				var ref Ref
+				err := db.Table(models.TableNameReferenceObject).
+					Where(SQLWhereGetByAllocationTxAndPath, allocationID, p).
```
lpoli, line 49: I suppose it requires parent path for the query. Please fix this.
Also this NewRefWalker struct is complex for other developers. Previous code was easier to understand.
Performance wise its same as previous.

code/go/0chain.net/blobbercore/reference/ref_walker.go
lpoli: Since it is static language and we have many tests written I don't think we have to become too cautious .  Returning rootPath for RefWalker typed nil value when Name() is called seems issue to me.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -293,7 +291,9 @@ func (fr *Ref) CalculateFileHash(ctx context.Context, saveToDB bool) (string, er
 func (r *Ref) CalculateDirHash(ctx context.Context, saveToDB bool) (string, error) {
 	// empty directory, return hash directly
 	if len(r.Children) == 0 && !r.childrenLoaded {
-		return r.Hash, nil
+		// skip empty directory in hash
+		return "", nil
```
lpoli, line 296: Need to be careful here. We can upload file to non-existing directories. For eg: `/d1/d2/d3/d4/file.txt`. So in this case if d1, d2, d3 and d4 did not exist then it will be created and it will have hash.
Later on file.txt can be deleted and directories can still have hash.

If returning empty string won't cause any issues then its good to go otherwise it is a bug.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -304,25 +304,24 @@ func (r *Ref) CalculateDirHash(ctx context.Context, saveToDB bool) (string, erro
...
-	for index, childRef := range r.Children {
-		childHashes[index] = childRef.Hash
-		childPathHashes[index] = childRef.PathHash
-		refNumBlocks += childRef.NumBlocks
-		size += childRef.Size
-	}
-	// fmt.Println("ref name and path, hash :" + r.Name + " " + r.Path + " " + r.Hash)
-	// fmt.Println("ref hash data: " + strings.Join(childHashes, ":"))
+	for _, childRef := range r.Children {
+		if len(childRef.Hash) > 0 {
```
lpoli, line 314: So here I see that if a child is empty directory then hash will equal to 0.

Again we need to be careful as it will impact whole system.

code/go/0chain.net/blobbercore/handler/directory.go
```diff
@@ -1,11 +1,81 @@
...
+				return errors.ThrowLog(err.Error(), constants.ErrBadDatabaseOperation, name)
+			}
+
+			// doesn't exists, create it, and check its parent
+
+			dir := reference.NewDirectoryRef()
+
+			dir.ActualFileSize = 0
+			dir.AllocationID = ctx.AllocationTx
+			dir.MerkleRoot = ""
```
lpoli, line 44: Why initialize with default value? It can be removed along with other lines

code/go/0chain.net/blobbercore/handler/directory.go
```diff
@@ -1,11 +1,81 @@
...
 
+import (
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/reference"
+	"github.com/0chain/errors"
+	"github.com/0chain/gosdk/constants"
+	"gorm.io/gorm"
+)
+
 // CreateDir is the handler to respond to create dir for allocation
 func CreateDir(ctx *Context) (interface{}, error) {
```
lpoli, line 11: My only concern over this function is: Do we only need directory info as reference_object?

I had previously checked how file was stored in a directory. I think file's hash was used by dividing it to 3,3,3,7 characters to store file.

lpoli (changes requested):
Empty dir hash has been set to empty string. I've commented about it. Please check.
Also this RefWalker struct is a bit complex. Its easy for another developer to make error by using this structure. Previous code was easier to understand.

Being extra cautious seemed redundant to me. It would be better to return error rather than to provide root path which might be future bug.

lpoli:
> > I suppose by auto-migration we mean adding/updating table schema, adding/updating indexes, etc. Also the migration code has not taken data migration into consideration so this is related to table schema only.
> > If that can be achieved by gorm fields then it is fine but if we need to create newer scripts then I don't think it is easier approach.
> > Also please add appropriate files changes under some commit. This commit is related to auto-migration of postgresql schema but has other changes as well.
> 
> This PR is submitted for `CreateDir`. Other changes are developed for it too. if we split them into other PRs again, our code will still keep smell. I would suggest to improve it pice by pice when we are fixing bugs or adding feature if it is possible.
> 
> It is a big challenge if someone assign me an issue to refactor all db code at the same time. It is real difficult to get it done in time. And we have to stop to work on any features /bugs .

It can be under same PR but I think under auto-migration commit message there were other changes that was irrelevant to auto-migration.

code/go/0chain.net/blobbercore/models/migration.go
lpoli: Then what is the point of checking if developer will put correct one?

code/go/0chain.net/blobbercore/models/migration.go
lpoli: So do you mean that by automation scripts we are only supplementing migration that GORM would otherwise be unable to migrate?

btw I did not get your first point. Can you please explain it?

code/go/0chain.net/blobbercore/reference/ref_walker.go
cnlangzi: app should be `Robustness`. it mean we can't leave any crash chance to consumer. eg

have a think what happens if it is removed?
```
var refwalker *RefWalker

ref.Name()
```

code/go/0chain.net/blobbercore/models/migration.go
cnlangzi: you can think there are 3 independent versions. Actually, they are .

code/go/0chain.net/blobbercore/models/migration.go
cnlangzi: try to test it on local  gorm.AutoMigrate(&ReferenceObject{}). you will see what happens.

code/go/0chain.net/blobbercore/reference/referencepath.go
cnlangzi: No. there are two issues on old code for some cases. just no guy to test and fix it. do you have a check on what the sql it generated is?

it is generated like `select * from reference_objects where (((parent_path='/d1/d2/d3/d4/d5' and allocation_id='cff048aaa3bdd858d0f52bfe833ef441464d79089e2d44c6272c3f8a1f50977c) or (prarent_path='/d1/d2/d3/d4' and allocation_id='cff048aaa3bdd858d0f52bfe833ef441464d79089e2d44c6272c3f8a1f50977c')) or.......) and (path=/d1/d2/d3/d4/d5/d6' and allocation_id='cff048aaa3bdd858d0f52bfe833ef441464d79089e2d44c6272c3f8a1f50977c' order by level,paht.

-  it will hit `max_expr_depth` or `max_stack_depth` issue on postgres if some guy has a big ref tree(eg. depth is 100). And his allocation will never work again. Because getRootRef might stop to work. It will kill almost of commands.
- i don't think too many `Or`  in where is not a performance issue.

code/go/0chain.net/blobbercore/reference/ref.go
cnlangzi: it hides empty dir in ref's Hash. if not, we hit the lock issue again. that is why it is changed. so we can free feel to create /delete empty dir as much as we want. if not, you need calculate hash , and send transaction to blockchain , and wait for verification form sharder.

code/go/0chain.net/blobbercore/reference/ref.go
cnlangzi: It is opposite to break system, it does compatible with it. empty dir will not effect old hash, it is invisible for it.

code/go/0chain.net/blobbercore/handler/directory.go
cnlangzi: yes , we need it in db. if not, zbox list will not work.

code/go/0chain.net/validatorcore/storage/models.go
cnlangzi: empty dir

code/go/0chain.net/blobbercore/datastore/postgres.go
lpoli: Yes that's what I mean.
We must use transaction for write operations. While for read operations we want to skip it.

code/go/0chain.net/blobbercore/reference/referencepath.go
lpoli: This function is required to calculate new Hashes when some file changes.
It takes all refs under subdirectories and gradually calculates hashes. Previous query worked with "parent_path" in query and now you are just querying with "path". I think this is a bug.

code/go/0chain.net/blobbercore/reference/ref.go
lpoli: Consider a filepath `/d1/d2/d3/d4/file.txt` and `file.txt` is deleted. Of course, d4 ref will have empty hash but what happens to d1, d2 and d3 ref?

code/go/0chain.net/blobbercore/handler/directory.go
lpoli: You misunderstood my concern. But its fine.

code/go/0chain.net/blobbercore/reference/ref_walker.go
cnlangzi: hi guys, do you know `Dbc`?  [Design by contract](https://en.wikipedia.org/wiki/Design_by_contract)

You must assert any input without valid value.if not, static check fails. Your app never crash by nil input like `nInvalid memory address or nil pointer dereference in golang`, `NullReferenceException in c#` and ` java.lang.NullPointerException in Java`

In my opinion, it is a basic rule to prevent app crash by nil/null point on coding.

code/go/0chain.net/blobbercore/reference/referencepath.go
cnlangzi: No, that is not true. please have check it again. it works similar old logic, just split them into one dir by one dir instead of getting them in a query, and from root to last leaf .

- Top() will jump to root dir
- loop Forward() till done. all dir will be loaded if it is necessary. if not, we can stop and return anytime.

code/go/0chain.net/blobbercore/reference/ref.go
cnlangzi: good question. I think `Hash` should be updated with "".  Because all nodes are re-calculated when root is calculated. And all changes are saved with root.  And I will test it again to verify it.

code/go/0chain.net/blobbercore/reference/ref_walker.go
lpoli: I fairly checked the above link. It suggests that client and supplier are obliged to fulfill certain preconditions and postconditions. Preconditions are to be fulfilled by client that is imposed by supplier and Postconditions are fulfilled by supplier that is imposed by client.

In our case we have to have precondition to not allow nil value and return error rather than let application crash.
My point was, for a nil value RefWalker it defaulted to rootpath "/", which might be issue later on.

Sriep:
Why do we need the auto migration changes coupled to the `createdir` changes? 

I find it often useful to initially make a compound PR with all the changes I want. Once completed I can usually see the best way to split the PR up into disjoint subsets.

code/go/0chain.net/blobbercore/allocation/dao.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	alloc := &models.Allocation{}
+	result := db.Table(models.TableNameAllocation).Where(SQLWhereGetByTx, allocationTx).First(alloc)
+
+	if result.Error == nil {
+		return alloc, nil
+	}
+
+	if !errors.Is(result.Error, gorm.ErrRecordNotFound) {
```
kushthedude, line 28: Can you move the dao integration & revamping to a separate PR?

code/go/0chain.net/blobbercore/handler/context.go
```diff
@@ -0,0 +1,129 @@
...
+
+// WithAuth verify alloation and signature
+func WithAuth(r *http.Request) (*Context, error) {
+
+	ctx := &Context{
+		Context: context.TODO(),
+		Request: r,
+		Store:   datastore.GetStore(),
+	}
+
```
kushthedude, line 91: This is completely unrelated to the create-dir but is a more generic implementation of the HTTP request/response manipulations. This is not required since gRPC code will already override this hacks for json response conversion.

code/go/0chain.net/blobbercore/models/migration.go
```diff
@@ -0,0 +1,109 @@
...
+	// Version format is "[table].[index].[column]"
+	//  + increase table version if any table is added or deleted
+	//  + increase index version if any index is changed
+	//  + increase column version if any column/constraint is changed
+	Version   string    `gorm:"column:version;primary_key"`
+	CreatedAt time.Time `gorm:"column:created_at"`
+	Scripts   []string  `gorm:"-"`
+}
+
+func (m *Migration) After(latest *Migration) bool {
```
kushthedude, line 27: Why are we adding models implementation & migration thing in the same PR which is already more than a monolith?

kushthedude (changes requested):
This PR not only re-implements the createDir but revamps the core architecture of how the handlers have been written & how we communicate with the db object at the moment.

My suggestions here would be to separate this monolithic PR into three small PRs:
- Addition of dao & auto-migrate
- Revamping the createDir handler
- Revamping the reference handlers

Functions which are related for interacting with the context to populate the request headers are completely unnecessary as once WASM is up we will deprecate this old handler & move to gRPC handlers which are present in the code. A point to note over here is, gRPC handlers are already revamping the context & headers logic by fetching the headers dynamically from request & passing into context.

code/go/0chain.net/blobbercore/reference/referencepath.go
lpoli: That I can see. But you are retrieving refs of some path while we need direct child refs of some path.

PR #409, opened 2021-10-28
:bug: Fixing issue with Encrypted Files

moldis:
The issue happens when the user wants to download his own encrypted file. In the current code it going to the wrong algorithm with shared files, this should be avoided as users must have the possibility to download their own files without auth_ticket.


cnlangzi (changes requested):
`encryption header` need to be deleted for both  owner and clients. so i think it should don't work for owner from zboxcli now. please try it on zbox

cnlangzi (approved):
LGTM. please update gosdk with latest commit with your gosdk PR too.

PR #422, opened 2021-11-07
move herumi installation to base image and use it to the entire docke…

rizary:
Move all herumi installation to base image in docker.local folder and change all dockerfile to use base_image tag instead of go image.

I have tested it locally and it works fine. This PR happened when I try to debug bls, so I create another branch to minimize changes.


cnlangzi (changes requested):
please update it on
- related bash scripts on ./docker.local/bin 
- related github actions

rizary:
@cnlangzi I have double check again locally and in the actions. As we can see in https://github.com/0chain/blobber/runs/4130997442?check_suite_focus=true#step:4:230 that `From build_base as blobber` works fine as expected.

cnlangzi:
> https://github.com/0chain/blobber/runs/4130997442?check_suite_focus=true#step:4:230

for example on https://github.com/0chain/blobber/blob/6e76d29148e1426d8001cf18da91c2e0a5974b94/.github/workflows/build-%26-publish-docker-image.yml#L80

it doesn't work. build_base was added by me. I only did it on 
- ./docker/bin/build.base.sh 
- ./docker/bin/build.blobber.sh
- ./docker/bin/build.validator.sh
- .github/workflows/dockerizer.yml

for now, you updated it on all Dockerfile. so you have to update related scripts and github actions.

cnlangzi (dismissed):
LGTM

PR #423, opened 2021-11-08
fix TestSignatureVerify with the updated gosdk

rizary:
I have updated TestSignatureVerify with the new bls from gosdk. This PR should be merged after this PR https://github.com/0chain/gosdk/pull/290 is merged and we pass all the test.

In order to test the code locally, you can uncomment the `replace ...` in the `go.mod`.


cnlangzi:
it is fixed on staging

PR #425, opened 2021-11-10
fix: rollback to herumi bls

cnlangzi:
It blocks everything now. so it is better to rollback to herumi bls first, and migrate it to miracl bls in new PR


rizary (dismissed):
LGTM

cnlangzi:
get this PR merged first to fix staging issue that block whole team.  gRPC tests will be fixed in a new PR.

PR #427, opened 2021-11-10
Trigger system tests from blobber PR

stewartie4:
This PR cannot merge unless system tests (which can be triggered here: https://github.com/0chain/blobber/actions/workflows/system_tests.yml )
pass


cnlangzi (changes requested):
@stewartie4  please move it to [Tests.yml](https://github.com/0chain/blobber/blob/523d8dc2426b85fc1292b5ef12b3682c2743aa70/.github/workflows/tests.yml#L32). and it should wait for `intergration_test` to test passed first. It will save us a lot of time to run `system_test` on invalid PR.

stewartie4:
@cnlangzi these don't run automatically, they run manually so the dev can choose to run these after the integration tests pass

cnlangzi (approved):

PR #429, opened 2021-11-12
Fix/integration tests GitHub

cnlangzi:
- run blobber outside docker on integration-tests. and save time to build docker images. It takes only 2-5 minutes now. it was 20-30 minutes.
- add commands `make local-init` `make local-build` and `make local-run` on Makefile for local development and debugging
- fixed some bugs on grpc test code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,17 +42,18 @@ jobs:
       - name: Clone Blobber
         uses: actions/checkout@v1
 
-      - name: Init Postgres
+      - name: Install Postgres
         run: |
           ./docker.local/bin/blobber.init.setup.sh
           cd docker.local/blobber1
           ../bin/blobber.start_github.sh
           docker ps
-      - name: Run Blobber
-        run: make local-run &
+
 
       - name: Run Tests
         run: |
+          make local-run &
+          sleep 10
           go17=$(which go)
           sudo CGO_ENABLED=1 $go17 test -tags bn256  ./... -args integration
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,14 +48,16 @@ jobs:
           cd docker.local/blobber1
           ../bin/blobber.start_github.sh
           docker ps
-
+      - name: Run Blobber
+        run: |
+          make local-run &
+          sleep 5
 
       - name: Run Tests
         run: |
-          make local-run &
-          sleep 10
           go17=$(which go)
-          sudo CGO_ENABLED=1 $go17 test -tags bn256  ./... -args integration
+          root=$(pwd)
+          sudo CGO_ENABLED=1 root=$root integration=1  $go17 test -tags bn256  ./...
 
   golangci:
     name: lint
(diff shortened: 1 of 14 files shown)


docker.local/b0docker-compose-github.yml
```diff
@@ -26,26 +26,6 @@ services:
...
-      - ./keys_config:/blobber/keysconfig
-      - ./blobber${BLOBBER}/data/tmp:/tmp
-    ports:
-      - 505${BLOBBER}:505${BLOBBER}
-      - 3150${BLOBBER}:3150${BLOBBER}
-    command: ./bin/blobber --port 505${BLOBBER} --grpc_port 3150${BLOBBER} --hostname 198.18.0.9${BLOBBER} --deployment_mode 0 --keys_file keysconfig/b0bnode${BLOBBER}_keys.txt --files_dir /blobber/files --log_dir /blobber/log --db_dir /blobber/data --minio_file keysconfig/minio_config.txt
-    networks:
-      default:
-      testnet0:
-        ipv4_address: 198.18.0.9${BLOBBER}
```
kushthedude, line 47: Why did you remove this?

Makefile
```diff
@@ -22,10 +24,44 @@ lint:
...
+	--port 5051 \
+	--grpc_port 31501 \
+	--hostname 127.0.0.1 \
+	--deployment_mode 0 \
+	--keys_file ../docker.local/keys_config/b0bnode1_keys.txt  \
+	--files_dir ./data/blobber/files \
+	--log_dir ./data/blobber/log \
+	--db_dir ./data/blobber/data  \
+	--minio_file ../docker.local/keys_config/minio_config.txt \
+	--config_dir ./data/blobber/config
```
kushthedude, line 63: Nice 👍🏽

.github/workflows/tests.yml
```diff
@@ -47,23 +42,29 @@ jobs:
...
         run: |
+          make local-build
+          sudo make local-run &
+          sleep 15
+          running=0
+          echo "checking blobber server status"
+          if [ $running -eq 0 ]; then
+            echo "checking blobber server status"
+            curl 127.0.0.1:5051 && $running=1 || sleep 2
+          fi
```
kushthedude, line 61: I would prefer having a job to test the dockerised builds as this is what would actually get deployed on production.

docker.local/b0docker-compose-github.yml
cnlangzi: it is implemented by `make local-run` instead .

cnlangzi: see https://github.com/0chain/blobber/blob/162c29b0f2b938160004354dfb770fdc7bcd235c/Makefile#L51

cnlangzi: only postgres and postgres-post are running inside docker. It is difficult to debug test code inside docker. so move them out from docker

.github/workflows/tests.yml
cnlangzi: No. it is not deployed on production . it is only working on github action. docker image is used on production.

cnlangzi: And it is not a dockerize build. it is `go build` and `run local app` in background . so check port to verify http server if it is up. if not, tests code might not work.

PR #446, opened 2021-11-26
Create dir logic added if given path is not available.

### LINKED ISSUE #401: Copy to root doesn’t seem to work if there are no folders
https://github.com/0chain/zboxcli/issues/67

princeparmar:
in copy command if given path is not available need to create whole tree and then copy that file.

Signed-off-by: Pradip Parmar <pradip@alfa-creator.com>

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/object_operation_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/object_operation_handler.go
@@ -914,15 +914,6 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
 		return nil, common.NewError("invalid_parameters", "Invalid connection id passed")
 	}
 
-	connectionObj, err := allocation.GetAllocationChanges(ctx, connectionID, allocationID, clientID)
-	if err != nil {
-		return nil, common.NewError("meta_error", "Error reading metadata for connection")
-	}
-
-	mutex := lock.GetMutex(connectionObj.TableName(), connectionID)
-	mutex.Lock()
-	defer mutex.Unlock()
-
 	objectRef, err := reference.GetReferenceFromLookupHash(ctx, allocationID, pathHash)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -1201,3 +1180,36 @@ func getExistingFileRef(fsh *StorageHandler, ctx context.Context, r *http.Reques
 	}
 	return nil
 }
+
+func createDir(ctx context.Context, connectionID, allocationID, clientID, dirPath string) error {
```
cnlangzi: hmm. it seems is not on right direction. I think we can add folders on RefTree only like '/'. and it can be saved in db.

cnlangzi (changes requested):
please check https://github.com/0chain/blobber/blob/7d4af78d867ea54b44e48aed2104df2038df2d85/code/go/0chain.net/blobbercore/reference/referencepath.go#L53 as example

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -935,10 +926,26 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
 	}
 
 	destRef, err = reference.GetReference(ctx, allocationID, destPath)
```
cnlangzi:
```suggestion
	destRef, err = reference.Mkdir(ctx, allocationID, destPath)
	if err != nil || destRef.Type != reference.DIRECTORY {
		return nil, common.NewError("invalid_parameters", "Invalid destination path. Should be a valid directory.")
	}
```

```
// Mkdir create dirs if they don't exits. do nothing if dir exists. last dir will be returned
func Mkdir(ctx context.Context, allocationID string, path string) (*Ref, error) {

	dirRef := &Ref{}

	dirs := strings.Split(path, "/")
	parentPath := ""
	currentPath := ""

	for _, dir := range dirs {

		currentPath += "/" + dir

		var ref *Ref

		err := db.Where(&Ref{AllocationID: allocationID, Path: currentPath}).First(ref).Error

		if err == nil {
			dirRef = ref
			parentPath = currentPath
			continue
		}

		// dir doesn't exists , create it
		if errors.Is(err, gorm.ErrRecordNotFound) {
			ref = &Ref{Type: DIRECTORY, AllocationID: allocationID, Name: dir, Path: currentPath, ParentPath: parentPath, PathLevel: 1}
			err = db.Create(ref).Error
			if err != nil {
				return nil, err
			}

			dirRef = ref
			parentPath = currentPath
			continue
		} else {
			// unexpected sql error
			return nil, err
		}
	}

	return dirRef, nil

}
```

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
cnlangzi: a simple demo code, please test it fully on local about WriteMarker issue

stewartie4:
could a system test be added for this new behaviour please?
https://github.com/0chain/system_test/blob/master/tests/cli_tests/zboxcli_file_copy_test.go
thanks

cnlangzi:
> could a system test be added for this new behaviour please? https://github.com/0chain/system_test/blob/master/tests/cli_tests/zboxcli_file_copy_test.go thanks

@princeparmar  please get `integration-tests` fixed first, and follow the [guideline](https://github.com/0chain/system_test#run-tests-against-an-existing-0chain-network-locally) to add system_test. thanks.

cnlangzi:
if you want to run system_test on github action, please use `blobber:merge` image

cnlangzi (approved):

stewartie4:
have system tests been ran/updated before this merge?

PR #458, opened 2021-12-08
fix cors and max_file_size checking

cnlangzi:
- enabled `OPTIONS` verb on some handlers
- ignored file size checking if max_file_size is 0
- rejected upload if actual size is greater than max_file_size.


kushthedude:
Why do we need options on some endpoints?

cnlangzi:
> Why do we need options on some endpoints?

cors header is checked via OPTION in preflight request. it will be blocked by browser if it is limited with POST/GET only.
see detail on https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

kushthedude (approved):

PR #460, opened 2021-12-09
Remove  max_file_size restriction

### LINKED ISSUE #459: Max file size logic should be removed
It is important to remove this field because as we know blobbers are independent nodes and therefore file won't be uploaded even though allocation has available space.

Consider we have 6 blobbers and if 3 blobbers set `max_file_size` to 10MB and other 3 blobbers set it to 1GB. New allocation can be created because we don't check for this field. But when user uploads, say >120 MB file, it will always fail.

It should be blobber responsibility to store arbitrary sizes as long as space for an allocation is available. Otherwise it will result in conflict among blobbers to store and not store files.

lpoli:
https://github.com/0chain/blobber/issues/459

Max file size restriction adds much more complexity and makes less sense to have.
Having said blobber to be able to store infinite size, `max_file_size` is not good to have. Upload will fail if few blobbers reject file because of file size restriction.

### DIFF
--- a/code/go/0chain.net/blobber/config.go
+++ b/code/go/0chain.net/blobber/config.go
@@ -51,7 +51,6 @@ func setupConfig() {
 	config.Configuration.MinioUseSSL = viper.GetBool("minio.use_ssl")
 
 	config.Configuration.Capacity = viper.GetInt64("capacity")
-	config.Configuration.MaxFileSize = viper.GetInt64("max_file_size")
 
 	config.Configuration.DBHost = viper.GetString("db.host")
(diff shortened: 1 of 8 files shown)


cnlangzi (changes requested):
please fix check issues, trigger 0chain system test by manually too. if not, the `merge` button is disabled.
I would suggest to fix ActualSize issue only. and leave max_file_size there. just configure it with 0 on 0chain_blobber.yml if unlimited size is required. 
let's waiting for @guruhubb  to confirm it.

cnlangzi (approved):
LGTM

PR #463, opened 2021-12-10
Feature/use system tests action

stewartie4:
Make system tests more user friendly:
- Kick off automatically on pull request commit against default staging values
- Manual option remains if you need to specify a custom branch/image 
- Manual system tests run no longer needs you to specify a blobber or validator image, instead it figures out what this image should be and tests against that

Example failure
image
Example success
image

### DIFF
--- a/.github/workflows/build-&-publish-docker-image.yml
+++ b/.github/workflows/build-&-publish-docker-image.yml
@@ -1,5 +1,9 @@
 name: Dockerize
 
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: true
+
 on:
   push: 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-&-publish-docker-image.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
   build_blobber:
     needs: build_base
-    runs-on: [self-hosted, build]
+    runs-on: [self-hosted, conductor-test]
```
stewartie4, line 49: revert once load-test is fixed

rizary (approved):
LGTM

PR #469, opened 2021-12-16
checkout current branch explicitly on manual system tests

stewartie4:
Existing workflow:
- built blobber base once then uploaded it 
- subsequent stages would download it 
- built blobber and validator then uploaded them
- subsequent stages would download and publish them

The uploading and downloading added minutes to the build time and was flaky when runners were rotated.
Amalgamated all of these stages into 2 - one for blobber and another for validator 

The result is some duplication (both build blobber base) but they now run much faster

- [x]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a branch at [0chain/system_test](https://github.com/0chain/system_test)


cnlangzi (approved):
LGTM

PR #495, opened 2022-01-11
Lp add authticket

lpoli:
GetRefs function was allowing only allocation owner to access refs.
It has been updated to allow access to repairer, collaborator and shared client.

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/object_operation_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/object_operation_handler.go
@@ -309,11 +309,6 @@ func (fsh *StorageHandler) DownloadFile(
 			payerID = clientID
 		}
 
-		if json.Unmarshal([]byte(authTokenString), &readmarker.AuthTicket{}) != nil {
-			return nil, common.NewErrorf("download_file",
-				"error parsing the auth ticket for download: %v", err)
-		}
-
 		// we only check content hash if its authticket is referring to a file
 		if authToken.RefType == zfileref.FILE && authToken.ActualFileHash != fileref.ActualFileHash {
 			return nil, errors.New("content hash does not match the requested file content hash")
--- a/code/go/0chain.net/blobbercore/handler/storage_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/storage_handler.go
@@ -769,7 +769,7 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
 	authTokenStr := r.FormValue("auth_token")
 	pathHash := r.FormValue("path_hash")
 
-	if !(path == "" || authTokenStr == "" || pathHash == "") {
+	if path == "" && authTokenStr == "" && pathHash == "" {
 		return nil, common.NewError("invalid_parameters", "empty path and authtoken")
 	}
 
--- a/code/go/0chain.net/blobbercore/handler/storage_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/storage_handler.go
@@ -889,17 +889,11 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
 		}
 	}
 
-	var latestRM *readmarker.ReadMarker
-	if latestRME, err := readmarker.GetLatestReadMarkerEntity(ctx, clientID); err == nil && latestRME != nil {
-		latestRM = latestRME.LatestRM
-	}
-
 	var refResult blobberhttp.RefResult
 	refResult.Refs = refs
 	refResult.TotalPages = totalPages
 	refResult.OffsetPath = newOffsetPath
 	refResult.OffsetDate = newOffsetDate
-	refResult.LatestRM = latestRM
 	if latestWM != nil {
 		refResult.LatestWM = &latestWM.WM
 	}
--- a/code/go/0chain.net/blobbercore/handler/storage_handler.go
+++ b/code/go/0chain.net/blobbercore/handler/storage_handler.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -54,6 +56,16 @@ func (fsh *StorageHandler) verifyAllocation(ctx context.Context, tx string,
 	}
 
 	return
+}
+
+func (fsh *StorageHandler) convertGormError(err error) error {
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return common.NewError("invalid_path", "path does not exist")
+		}
+		return common.NewError("db_error", err.Error())
+	}
+	return err
 }
 
 func (fsh *StorageHandler) verifyAuthTicket(ctx context.Context, authTokenString string, allocationObj *allocation.Allocation, refRequested *reference.Ref, clientID string) (bool, error) {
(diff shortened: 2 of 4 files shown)


code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -128,7 +128,7 @@ type PaginatedRef struct { //Gorm smart select fields.
 	UpdatedAt time.Time      `gorm:"column:updated_at" json:"updated_at,omitempty"`
 	DeletedAt gorm.DeletedAt `gorm:"column:deleted_at" json:"-"` // soft deletion
 
-	ChunkSize int64 `gorm:"column:chunk_size" dirlist:"chunk_size" filelist:"chunk_size"`
```
cnlangzi, line 131: why `dirList` and `fileList` should be update with `json`?  they are used in GetListingFieldsMap. `json` tag will not work.

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
 		return nil, common.NewError("invalid_parameters", "Invalid allocation id passed."+err.Error())
 	}
 
+	clientID := ctx.Value(constants.ContextKeyClient).(string)
+	if len(clientID) == 0 {
+		return nil, common.NewError("invalid_operation", "Client id is required")
+	}
+
+	publicKey, _ := ctx.Value(constants.ContextKeyClientKey).(string)
```
cnlangzi, line 750: is it better to says it is required like clientId? why client doesn't send it?

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
...
-	valid, err := verifySignatureFromRequest(allocationTx, clientSign, allocationObj.OwnerPublicKey)
+
+	valid, err := verifySignatureFromRequest(allocationTx, clientSign, publicKey)
 	if !valid || err != nil {
 		return nil, common.NewError("invalid_signature", "Invalid signature")
 	}
 
 	allocationID := allocationObj.ID
-	clientID := ctx.Value(constants.ContextKeyClient).(string)
-	if len(clientID) == 0 || allocationObj.OwnerID != clientID {
```
cnlangzi, line 741: does it break some methods(eg:`InsertShare`) if we don't check ownership? is it a security issue?

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
...
-	if len(path) == 0 {
-		return nil, common.NewError("invalid_parameters", "Invalid path")
+	authTokenStr := r.FormValue("auth_token")
+	pathHash := r.FormValue("path_hash")
+
+	if path == "" && authTokenStr == "" && pathHash == "" {
+		return nil, common.NewError("invalid_parameters", "empty path and authtoken")
 	}
 
+	var pathRef *reference.Ref
```
cnlangzi, line 776: `switch+case+fallthrough` is very difficult to understand how it works? could we extract them to 2 methods: `loadRefByPath` and `loadRefByAuthticket` ?

cnlangzi (dismissed):

code/go/0chain.net/blobbercore/reference/ref.go
lpoli: @cnlangzi The structure you are seeing is `PaginatedRef`. Only json tags is used in all fields.
You updated the `ChunkSize` field but without json tag it is not received in the client side

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: @cnlangzi Any suggestion?
Client may forget to send `client id`. Today we use gosdk which sends it by default. There may be other application that don't use gosdk( and have their own sdk).

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: @cnlangzi It is checked. But whole PR is about allowing access to other users as well i.e. collaborator, repairer, and shared users. That's why all three conditions are checked.

```
if clientID == allocationObj.OwnerID || clientID == allocationObj.RepairerID || reference.IsACollaborator(ctx, pathRef.ID, clientID) {
			break
		}
		if authTokenStr == "" {
			return nil, common.NewError("unauthorized_request", "client is not authorized for the requested resource")
		}
```

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: Is it difficult? It feels intuitive to me.

code/go/0chain.net/blobbercore/handler/storage_handler.go
cnlangzi: if he forget to send it, throw error may be better like you do in client_id

lpoli: do you mean we should instead return "invalid client id"?

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
 		return nil, common.NewError("invalid_parameters", "Invalid allocation id passed."+err.Error())
 	}
 
+	clientID := ctx.Value(constants.ContextKeyClient).(string)
+	if len(clientID) == 0 {
+		return nil, common.NewError("invalid_operation", "Client id is required")
+	}
+
+	publicKey, _ := ctx.Value(constants.ContextKeyClientKey).(string)
+	if publicKey == "" {
```
moldis, line 751: @lpoli consider to replace check like len(publicKey)  == 0

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -309,11 +309,6 @@ func (fsh *StorageHandler) DownloadFile(
 			payerID = clientID
 		}
 
-		if json.Unmarshal([]byte(authTokenString), &readmarker.AuthTicket{}) != nil {
```
moldis, line 312: @lpoli Why it was removed from here?

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
...
+		authToken := &readmarker.AuthTicket{}
+		if err := json.Unmarshal([]byte(authTokenStr), authToken); err != nil {
+			return nil, common.NewError("json_unmarshall_error", fmt.Sprintf("error parsing authticket: %v", authTokenStr))
+		}
+
+		shareInfo, err := reference.GetShareInfo(ctx, authToken.ClientID, authToken.FilePathHash)
+		if err != nil {
+			return nil, fsh.convertGormError(err)
+		}
+		if shareInfo.Revoked {
```
moldis, line 807: Besides Revoked, share info may have the time since it's available.

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -730,25 +742,100 @@ func (fsh *StorageHandler) GetRefs(ctx context.Context, r *http.Request) (*blobb
...
+			return nil, common.NewError("unauthorized_request", "client is not authorized for the requested resource")
+		}
+		fallthrough
+	case authTokenStr != "":
+		authToken := &readmarker.AuthTicket{}
+		if err := json.Unmarshal([]byte(authTokenStr), authToken); err != nil {
+			return nil, common.NewError("json_unmarshall_error", fmt.Sprintf("error parsing authticket: %v", authTokenStr))
+		}
+
+		shareInfo, err := reference.GetShareInfo(ctx, authToken.ClientID, authToken.FilePathHash)
```
moldis, line 803: Does it works with Directory?

moldis (changes requested):
Minor comments. The key question - does it work with current auth_ticket implementation?

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: https://stackoverflow.com/a/18595217/7661958

I took reference from above link

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: Because it is also checked in line 302.

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: Yes it does have time. But it is checked by `authToken.Verify` method.

code/go/0chain.net/blobbercore/handler/storage_handler.go
lpoli: Yes it does.

moldis (approved):
Looks good

PR #513, opened 2022-01-23
fix(node): added cli argument --hosturl to register url on zcn

cnlangzi:
- [x]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [x]  Merge your system tests PR to master BEFORE merging this PR

- zbox:
- zwalelet:
- gosdk:
- 0chain:
- system_test: 

changes:
- added cli argument `--hosturl` . it will be registered as blobber base url on blockchain if it is passed on cli

```
./blobber --hosturl https://dev.0chain.net/blobber1
```

### DIFF
--- a/code/go/0chain.net/blobber/flags.go
+++ b/code/go/0chain.net/blobber/flags.go
@@ -21,6 +21,7 @@ var (
 	httpsPort         int
 	httpsKeyFile      string
 	httpsCertFile     string
+	hostUrl           string
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


kushthedude (approved):

code/go/0chain.net/blobber/flags.go
```diff
@@ -32,12 +33,14 @@ func init() {
...
-	flag.IntVar(&httpsPort, "https-port", 0, "https-port")
-	flag.StringVar(&httpsCertFile, "https-cert-file", "", "https-cert-file")
-	flag.StringVar(&httpsKeyFile, "https-key-file", "", "https-key-file")
+	flag.IntVar(&httpsPort, "https_port", 0, "https_port")
+	flag.StringVar(&httpsCertFile, "https_cert_file", "", "https_cert_file")
+	flag.StringVar(&httpsKeyFile, "https_key_file", "", "https_key_file")
 	flag.StringVar(&hostname, "hostname", "", "hostname")
 	flag.StringVar(&configDir, "config_dir", "./config", "config_dir")
 
+	flag.StringVar(&hostUrl, "hosturl", "", "register url on blockchain instead of [schema://hostname+port] if it has value")
```
kushthedude, line 42: Can I provide both hosturl as well as hostname? It is contradicting to have hostname with port as well as hosturl, do we have a use-case for hosturl?

cnlangzi: the url is concat by `http(s)://hostname:port`. but it is https://dev.0chain.net/blobber01 in our networks

so it doesn't work with --hostname https://dev.0chain.net/blobber01. We don't know which pattern should be used. 

because hostname is not use-case, so just keep `hosturl` as `hostname`

PR #529, opened 2022-02-05
Issue 524

### LINKED ISSUE #524: Optimize Root Hash Calculation
Check this function [CalculateHash](https://github.com/0chain/blobber/blob/staging/code/go/0chain.net/blobbercore/reference/ref.go#L381). This function has too many inefficiencies and it can be the cause of postgres going to recovery mode frequently.

I think we should address this function optimization in the following order:
1. Consider having hash calculation in single `for` loop. Currently it is using two `for` loops which is redundant.
2. Consider reducing columns in sql queries. We don't need all columns to calculate hash
3. Consider updating only necessary columns
4. Consider ordering `refs` with `ref.Path` value rather than `ref.LookupHash`
5. Optimize any other inefficiencies on the road.

Have system test pass in each step. The change of ordering with `ref.Path` requires changes in `gosdk` as well.

After above changes:
```
Above changes is still not enough because it can query for million rows and if we consider row size 
after reducing columns in sql query to be `0.5KB` then it requires `500MB` RAM memory. And it is for 
a single root hash calculation. So we need to have queries in batch and have progressive hash calculation.
 ```
We need to also limit path length. It is critical to do so for the performance. This path limit task will be another task as it might requires changes in smart contract as well. 

For now this is simple and effective solution.

### PR DESCRIPTION
**Pull request by abouroumine**
Tests

- [ ]  Merge your system tests PR to master BEFORE merging this PR

- zbox: https://github.com/0chain/zboxcli/pull/192
- zwalelet:
- gosdk: https://github.com/0chain/gosdk/pull/416
- 0chain:
- system_test: https://github.com/0chain/system_test/pull/218

### CODE REVIEW

File: code/go/0chain.net/blobbercore/allocation/attributesfilechange.go [hunk 1]
```diff
@@ -27,7 +27,8 @@ func (ac *AttributesChange) ProcessChange(ctx context.Context, _ *AllocationChan
 	path = filepath.Clean(path)
 
 	// root reference
-	ref, err = reference.GetReferencePath(ctx, ac.AllocationID, ac.Path)
+	ref, err = reference.GetReferencePath2(ctx, ac.AllocationID, ac.Path)
```
moldis: Name of function seems wrong

code/go/0chain.net/blobbercore/allocation/file_changer_update.go
```diff
@@ -90,7 +92,7 @@ func (nf *UpdateFileChanger) ProcessChange(ctx context.Context, change *Allocati
 		return nil, common.NewErrorf("process_update_file_change",
 			"setting file attributes: %v", err)
 	}
-
+	fmt.Println("Calculate Hash Need To Update Data !!!")
```
moldis: This logs are not relevant here, please make it more detail or just remove

code/go/0chain.net/blobbercore/allocation/renamefilechange_test.go
```diff
@@ -159,9 +159,10 @@ func TestBlobberCore_RenameFile(t *testing.T) {
 			expectingError:  false,
 			setupDbMock: func() {
 				mocket.Catcher.Reset()
-
-				query := `SELECT * FROM "reference_objects" WHERE ("reference_objects"."allocation_id" = $1 AND "reference_objects"."path" = $2 OR (path LIKE $3 AND allocation_id = $4)) AND "reference_objects"."deleted_at" IS NULL ORDER BY level, lookup_hash%!!(string=allocation id)!(string=/old_dir/%!)(MISSING)!(string=/old_dir)(EXTRA string=allocation id)`
+				//query := `SELECT "id","allocation_id","type","name","path","parent_path","size","hash","path_hash","content_hash","merkle_root","actual_file_size","actual_file_hash","attributes","chunk_size","lookup_hash" FROM "reference_objects" WHERE ("reference_objects"."allocation_id" = $1 AND "reference_objects"."path" = $2 OR (path LIKE $3 AND allocation_id = $4)) AND "reference_objects"."deleted_at" IS NULL ORDER BY level, path%!!(string=allocation id)!(string=/old_dir/%!)(MISSING)!(string=/old_dir)(EXTRA string=allocation id)`
```
moldis: No comments in the code, especially in PR for review

code/go/0chain.net/blobbercore/handler/file_command_add.go
```diff
@@ -36,7 +36,8 @@ func (cmd *AddFileCommand) IsAuthorized(ctx context.Context, req *http.Request,
 		return common.NewError("invalid_parameters",
 			"Invalid parameters. Error parsing the meta data for upload."+err.Error())
 	}
-	exisitingFileRef, _ := reference.GetReference(ctx, allocationObj.ID, fileChanger.Path)
+	// Update GetReference to GetReferenceID
+	exisitingFileRef, _ := reference.GetReferenceID(ctx, allocationObj.ID, fileChanger.Path)
```
lpoli: I think its better to atleast log error if any. In any case of err other than `gorm.ErrRecordNotFound` we should log actual error and probably return `common.NewError("some_code", "database error")

lpoli: @abouroumine 
I am not sure about using GetReferenceID. Have you checked it with `ProcessContent` function as well ?
In other `file_command*` code `cmd.existingFileRef` is used which infers that we might need to consider using `reference.GetReference`.

Above all, I think its best to have `single responsibility principle`. That's why `cmd.existingFileRef` should have been assigned before calling `IsAuthorize`.

Please make changes for `IsAuthorize` function to have single responsibility.

code/go/0chain.net/blobbercore/handler/file_command_delete.go
```diff
@@ -28,6 +28,7 @@ func (cmd *FileCommandDelete) IsAuthorized(ctx context.Context, req *http.Reques
 	if path == "" {
 		return common.NewError("invalid_parameters", "Invalid path")
 	}
+	// Update GetReference to GetReferenceID
 	cmd.exisitingFileRef, _ = reference.GetReference(ctx, allocationObj.ID, path)
```
lpoli: https://github.com/0chain/blobber/pull/529/files#r816421915

code/go/0chain.net/blobbercore/handler/file_command_update.go
```diff
@@ -42,7 +42,7 @@ func (cmd *UpdateFileCommand) IsAuthorized(ctx context.Context, req *http.Reques
 	if cmd.fileChanger.ChunkSize <= 0 {
 		cmd.fileChanger.ChunkSize = fileref.CHUNK_SIZE
 	}
-
+	// Update GetReference to GetReferenceID
 	cmd.exisitingFileRef, _ = reference.GetReference(ctx, allocationObj.ID, cmd.fileChanger.Path)
```
lpoli: https://github.com/0chain/blobber/pull/529/files#r816421915

code/go/0chain.net/blobbercore/handler/handler.go
```diff
@@ -64,7 +64,7 @@ func SetupHandlers(r *mux.Router) {
 	r.HandleFunc("/v1/file/objectpath/{allocation}", common.ToJSONResponse(WithReadOnlyConnection(ObjectPathHandler)))
 	r.HandleFunc("/v1/file/referencepath/{allocation}", common.ToJSONResponse(WithReadOnlyConnection(ReferencePathHandler)))
 	r.HandleFunc("/v1/file/objecttree/{allocation}", common.ToJSONResponse(WithReadOnlyConnection(ObjectTreeHandler)))
-	r.HandleFunc("/v1/file/refs/{allocation}", common.ToJSONResponse(WithReadOnlyConnection(RefsHandler))).Methods(http.MethodGet, http.MethodOptions)
+	r.HandleFunc("/v1/file/refs/{allocation}", common.ToJSONResponse(WithReadOnlyConnection(RefsHandler))).Methods(http.MethodGet, http.MethodOptions) // no db optimization yet.
```
lpoli: Please check for this comment you have left

config/0chain_blobber.yaml
```diff
@@ -41,7 +41,7 @@ write_lock_timeout: 1m
 update_allocations_interval: 1m
 
 # delegate wallet (must be set)
-delegate_wallet: '2f34516ed8c567089b7b5572b12950db34a62a07e16770da14b15b170d0d60a9'
+delegate_wallet: 'ffb5f45e78cefc18e5b1c3fcdc9b0c87450dedc38d48abc7589e8f3af41a56c2'
```
lpoli: I think we don't need this change. You may have done this for manual testing.

config/0chain_validator.yaml
```diff
@@ -1,7 +1,7 @@
 version: 1.0
 
 # delegate wallet (must be set)
-delegate_wallet: '86b147de5be951a5ab0c6b1732596686185c5e11c3bdf76a112183f828f39ba1'
+delegate_wallet: 'ffb5f45e78cefc18e5b1c3fcdc9b0c87450dedc38d48abc7589e8f3af41a56c2'
```
lpoli: https://github.com/0chain/blobber/pull/529/files#r816429110

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -474,18 +472,16 @@ func (fsh *StorageHandler) CommitWrite(ctx context.Context, r *http.Request) (*b
 	if err := writePreRedeem(ctx, allocationObj, &writeMarker, clientIDForWriteRedeem); err != nil {
 		return nil, err
 	}
-
 	err = connectionObj.ApplyChanges(ctx, writeMarker.AllocationRoot)
 	if err != nil {
 		return nil, err
 	}
-	rootRef, err := reference.GetReference(ctx, allocationID, "/")
+	rootRef, err := reference.GetReferenceHash(ctx, allocationID, "/")
```
lpoli: @abouroumine `connectionObj.ApplyChanges` can also return same rootHash. Can we do that?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -693,7 +690,8 @@ func (fsh *StorageHandler) UpdateObjectAttributes(ctx context.Context, r *http.R
 	defer mutex.Unlock()
 
 	var ref *reference.Ref
-	ref, err = reference.GetReferenceFromLookupHash(ctx, alloc.ID, pathHash)
+	// Update GetReferenceFromLookupHash to GetOnlyReferencePathFromLookupHash
+	ref, err = reference.GetOnlyReferencePathFromLookupHash(ctx, alloc.ID, pathHash)
```
lpoli: Would `GetReferencePathFromLookupHash` be better than `GetOnlyReferencePathFromLookupHash`?

lpoli: Would function name `GetReferencePathFromLookupHash` be better than `GetOnlyReferencePathFromLookupHash`?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -822,7 +822,8 @@ func (fsh *StorageHandler) DeleteFile(ctx context.Context, r *http.Request, conn
 		return nil, common.NewError("invalid_parameters", "Invalid path")
 	}
 
-	fileRef, _ := reference.GetReference(ctx, connectionObj.AllocationID, path)
+	//fileRef, _ := reference.GetReference(ctx, connectionObj.AllocationID, path)
+	fileRef, _ := reference.GetReferenceDelete(ctx, connectionObj.AllocationID, path)
```
lpoli: @abouroumine Please don't ignore errors. Atleast log it.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -822,7 +822,8 @@ func (fsh *StorageHandler) DeleteFile(ctx context.Context, r *http.Request, conn
...
 	_ = ctx.Value(constants.ContextKeyClientKey).(string)
```
lpoli, line 822: Remove this redundant line#827 as well

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -875,7 +875,8 @@ func (fsh *StorageHandler) CreateDir(ctx context.Context, r *http.Request) (*blo
 		return nil, common.NewError("invalid_parameters", "Invalid dir path passed")
 	}
 
-	exisitingRef := fsh.checkIfFileAlreadyExists(ctx, allocationID, dirPath)
+	// Update checkIfFileAlreadyExists to checkIfFileRefIDAlreadyExists
+	exisitingRef := fsh.checkIfFileRefIDAlreadyExists(ctx, allocationID, dirPath)
```
lpoli: We can directly call `reference.GetReferenceID(ctx, allocationID, path)`. No need to call `checkIfFileRefIDAlreadyExists`.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -910,9 +911,9 @@ func (fsh *StorageHandler) CreateDir(ctx context.Context, r *http.Request) (*blo
 	formData.ConnectionID = connectionID
 	formData.ActualHash = "-"
 	formData.ActualSize = 1
-
+	// We Need to do a Proper Look here
```
lpoli: @abouroumine What about this comment?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -1035,14 +1037,27 @@ func getFileOperation(r *http.Request) string {
 	return mode
 }
 
-func getExistingFileRef(fsh *StorageHandler, ctx context.Context, r *http.Request, allocationObj *allocation.Allocation, fileOperation string) *reference.Ref {
```
lpoli, line 1048: Please Delete commented code.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -1035,14 +1037,27 @@ func getFileOperation(r *http.Request) string {
...
+//		}
+//	}
+//	return nil
+//}
+
+func getExistingFileRefID(fsh *StorageHandler, ctx context.Context, r *http.Request, allocationObj *allocation.Allocation, fileOperation string) *reference.Ref {
 	if fileOperation == constants.FileOperationInsert || fileOperation == constants.FileOperationUpdate {
 		var formData allocation.UpdateFileChanger
 		uploadMetaString := r.FormValue(getFormFieldName(fileOperation))
 		err := json.Unmarshal([]byte(uploadMetaString), &formData)
```
lpoli: @abouroumine If err is not nil then it is better to return error rather than fileRef being nil.
I think its better to add `error` in return and return this error to the client

code/go/0chain.net/blobbercore/reference/objectpath.go
```diff
@@ -19,6 +19,7 @@ type ObjectPath struct {
 
 // TODO needs to be refactored, current implementation can probably be heavily simplified
 func GetObjectPath(ctx context.Context, allocationID string, blockNum int64) (*ObjectPath, error) {
+	// Update GetRefWithSortedChildren to GetRefWithSortedChildrenForObjectPath
```
lpoli: @abouroumine Address this comment

code/go/0chain.net/blobbercore/reference/objectpath.go
```diff
@@ -66,6 +67,7 @@ func GetObjectPath(ctx context.Context, allocationID string, blockNum int64) (*O
 				curRef = child
 				break
 			}
+			// Update GetRefWithSortedChildren to GetRefWithSortedChildrenForObjectPath
```
lpoli: This as well

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -280,7 +381,8 @@ func GetRefWithSortedChildren(ctx context.Context, allocationID, path string) (*
 	var refs []*Ref
 	db := datastore.GetStore().GetTransaction(ctx)
 	db = db.Where(Ref{ParentPath: path, AllocationID: allocationID}).Or(Ref{Type: DIRECTORY, Path: path, AllocationID: allocationID})
-	err := db.Order("level, lookup_hash").Find(&refs).Error
+	//err := db.Order("level, lookup_hash").Find(&refs).Error
+	err := db.Order("level, path").Find(&refs).Error
```
lpoli: @abouroumine I think its more appropriate to do `db.Order("path, level")`.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -324,55 +426,55 @@ func (fr *Ref) GetFileHashData() string {
...
 	return fr.Hash, err
 }
 
 func (r *Ref) CalculateDirHash(ctx context.Context, saveToDB bool) (string, error) {
 	// empty directory, return hash directly
+
 	if len(r.Children) == 0 && !r.childrenLoaded {
 		return r.Hash, nil
 	}
 	sort.SliceStable(r.Children, func(i, j int) bool {
```
lpoli: @abouroumine I think if we do `db.Order("path", "level")` then this sorting is redundant. Do you think so? 
Also the comparison it makes with `LookupHash` is questionable. We cannot order path with lookuphashes

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -422,6 +524,80 @@ func DeleteReference(ctx context.Context, refID int64, pathHash string) error {
 	return db.Where("path_hash = ?", pathHash).Delete(&Ref{ID: refID}).Error
 }
 
+func (r *Ref) SaveFile(ctx context.Context) error {
```
lpoli: @abouroumine Please rename this function to something like `SaveFileRef` or `SaveFileRow`. Also same for `SaveDir`

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -324,55 +426,55 @@ func (fr *Ref) GetFileHashData() string {
...
-		if err != nil {
-			return "", err
-		}
-	}
 	childHashes := make([]string, len(r.Children))
 	childPathHashes := make([]string, len(r.Children))
 	var refNumBlocks int64
 	var size int64
 	for index, childRef := range r.Children {
+		_, err := childRef.CalculateHash(ctx, saveToDB)
```
lpoli: @abouroumine I think `childRef.CalculateHash` is required only for affected refs. For unaffected refs we can simply get hash value from `Hash` field.

We should avoid hash calculation where we can because it is computationally resource consuming.
It would be great to distinguish affected and unaffected rows. For example if path is `/a/b/c/d/e.txt` then hash calculation is basically required for refs with paths `/`, `/a`, `/a/b`, `/a/b/c`, `/a/b/c/d` and `/a/b/c/d/e.txt`. Correct me If I am wrong here. @abouroumine @moldis

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -22,6 +22,71 @@ func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, err
...
+		"lookup_hash", "thumbnail_hash")
+	pathsAdded := make(map[string]bool)
+	for _, path := range paths {
+		path = strings.TrimSuffix(path, "/")
+		if _, ok := pathsAdded[path]; !ok {
+			db = db.Where(Ref{ParentPath: path, AllocationID: allocationID})
+			pathsAdded[path] = true
+		}
+		depth := len(GetSubDirsFromPath(path)) + 1
+		curPath := filepath.Dir(path)
```
lpoli: This `curPath` prevails from previous code. I kind of don't like this logic. We could simply split path with `/` and get parent path by joining elements.
Something like:
```
splittedPaths := strings.Split("path", "/")
for i:=0;i<len(splittedPaths);i++{
    newPath := strings.Join(splittedPaths[0:i], "/")
}
```

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -22,6 +22,71 @@ func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, err
...
+				pathsAdded[curPath] = true
+			}
+			curPath = filepath.Dir(curPath)
+		}
+	}
+
+	// root reference_objects with parent_path=""
+	db = db.Or("parent_path = ? AND allocation_id = ?", "", allocationID)
+	//err := db.Order("level, lookup_hash").Find(&refs).Error
+	err := db.Order("level, path").Find(&refs).Error
```
lpoli: Note Order("path", "level")

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -22,6 +22,71 @@ func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, err
...
+	for i := 1; i < len(refs); i++ {
+		if _, ok := refMap[refs[i].ParentPath]; !ok {
+			return nil, common.NewError("invalid_dir_tree", "DB has invalid tree.")
+		}
+		if _, ok := refMap[refs[i].Path]; !ok {
+			refMap[refs[i].ParentPath].AddChild(&refs[i])
+			refMap[refs[i].Path] = &refs[i]
+		}
+	}
+	if _, err := refs[0].CalculateHash(ctx, false); err != nil {
```
lpoli, line 84: We may not need to calculate hash always when calling `GetReferenceForHashCalculationFromPaths`. Let's have a flag for whether to calculate hash or not.

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -105,12 +171,13 @@ func GetObjectTree(ctx context.Context, allocationID, path string) (*Ref, error)
...
-		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "/%"), allocationID)
+		db = db.Or("path LIKE ? AND allocation_id = ?", path+"/%", allocationID)
 	} else {
-		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "%"), allocationID)
+		db = db.Or("path LIKE ? AND allocation_id = ?", path+"%", allocationID)
 	}
 
-	err := db.Order("level, lookup_hash").Find(&refs).Error
+	//err := db.Order("level, lookup_hash").Find(&refs).Error
+	err := db.Order("level, path").Find(&refs).Error
```
lpoli: Note Order("path", "level")

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -124,8 +125,16 @@ func (fsh *StorageHandler) GetAllocationUpdateTicket(ctx context.Context, r *htt
...
-	fileReference, err := reference.GetReference(ctx, allocationID, path)
+//func (fsh *StorageHandler) checkIfFileAlreadyExists(ctx context.Context, allocationID, path string) *reference.Ref {
+//	fileReference, err := reference.GetReference(ctx, allocationID, path)
+//	if err != nil {
+//		return nil
+//	}
+//	return fileReference
+//}
+
+func (fsh *StorageHandler) checkIfFileRefIDAlreadyExists(ctx context.Context, allocationID, path string) *reference.Ref {
```
lpoli: Please put this function's body to previous function name

lpoli (changes requested):
@abouroumine 
Honestly I really appreciate your effort. I have commented in multiple places but as a sum up those changes are not much to work out.
Everyone has own method to solve problem. Though I would like to roughly explain my approach to this problem.

1. I would first create a postgres docker container and insert randomly generated millions rows with only necessary fields
2. Then calculate hashes and update all rows
3. After above step, make change in some row and re-calculate hash for the affected rows.
4. Calculate time taken so we can have performance difference.
5. Verify that calculation made by new function is correct.
6. Apply changes to blobber code base

code/go/0chain.net/blobbercore/allocation/attributesfilechange.go
abouroumine: This is a new function in code/go/0chain.net/blobbercore/reference/referencepath.go in line 25.

code/go/0chain.net/blobbercore/allocation/file_changer_update.go
abouroumine: indeed there are some left logs that I reused and were not removed yet.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
abouroumine: Both Functions are used in different places. The GetOnlyReferencePathFromLookupHash function selects only "id","path" where GetReferencePathFromLookupHash selects "id","name","path","hash","size","merkle_root". What I tried here is reduce the number of columns selected in Database query througth these two functions

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
abouroumine: The problems revolving around those comments already resolved. just forgot to remove comments.

code/go/0chain.net/blobbercore/reference/ref.go
abouroumine: I think you are right the sorting is indeed redundant. and can you please explain more the difference between db.Order("path","level") than db.Order("level","path").

code/go/0chain.net/blobbercore/reference/ref.go
abouroumine: Should we then introduce a new parameter that identifies the Affected and Unaffected Refs ? Especially if we are dealing with a file update / copy / move.

code/go/0chain.net/blobbercore/handler/file_command_add.go
abouroumine: I think unlike in file_command_update / file_command_delete, in file_command_add we are just checking if the File Ref exists. and we do not use it in ProcessContent. that is why GetReferenceID was used instead of GetReference.
Now that you have mentioned we can even reduce the columns selected for other file_command_* depending on the columns used for each command in each ProcessContent and other related functions.

code/go/0chain.net/blobbercore/allocation/attributesfilechange.go
lpoli: This is fine by me

code/go/0chain.net/blobbercore/handler/file_command_add.go
lpoli: @abouroumine Its fine that you've called `GetReferenceID`. My point is that it is called inside `IsAuthorized` function making it hard to debug later on. We can call for existingRef just before `IsAuthorized` function is called and check for `ref` availability.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: @abouroumine Okay then would it be better if we do same thing as you did with `GetReferencePath2`?
In this case, `GetReferencePathFromLookupHash2` and provide doc string in the function as well

code/go/0chain.net/blobbercore/reference/ref.go
lpoli: @abouroumine Now I think `level` is redundant. But the point was that sql ordering is heavy operation. First it gets all the rows and orders by some field and in that ordered result it orders again.
Also I think the difference between them is huge. In db.Order("path", "level") it first orders by path and then by level. In db.Order("level", "path") it first orders by level and then by path. 
In this two case they will have different set of ordered result.

`path` is unique and is sortable. `level` is not unique. So for example, `/a/b/c`, `/d/e/f` and `/d/e` are three paths.

If `db.Order('path', 'level')` then it results in:
/a/b/c
/d/e
/d/e/f

If `db.Order('level', 'path')` then it results in:
/d/e
/a/b/c
/d/e/f

code/go/0chain.net/blobbercore/reference/ref.go
lpoli: To identify affected rows new parameter is not required. Maybe it is required have a flag parameter to whether save or not save into database

code/go/0chain.net/blobbercore/reference/referencepath.go
abouroumine: this update may trigger errors and wrong paths if the paths include "." as parent_path. and give wrong paths for the DB request. for now I may keep it until I figure out why the paths are not similar.

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -105,12 +167,11 @@ func GetObjectTree(ctx context.Context, allocationID, path string) (*Ref, error)
...
 	if path != "/" {
-		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "/%"), allocationID)
+		db = db.Or("path LIKE ? AND allocation_id = ?", path+"/%", allocationID)
 	} else {
-		db = db.Or("path LIKE ? AND allocation_id = ?", (path + "%"), allocationID)
+		db = db.Or("path LIKE ? AND allocation_id = ?", path+"%", allocationID)
 	}
-
-	err := db.Order("level, lookup_hash").Find(&refs).Error
+	err := db.Order("path, level").Find(&refs).Error
```
lpoli, line 174: Since path are unique fields, lets remove `level` here.

lpoli: Since path are unique fields, lets remove `level` here.

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -88,14 +151,13 @@ func PathExists(ctx context.Context, allocationID, path string) (bool, error) {
 
 	var refs []Ref
 	db := datastore.GetStore().GetTransaction(ctx)
-
+	db = db.Select("id")
```
lpoli: About this `PathExists` function, there is `IsRefExists` function that I recently merged into staging. Lets replace this function call by `IsRefExists`. 
`IsRefExists` is more optimized than `PathExists`

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -46,11 +109,11 @@ func GetReferencePathFromPaths(ctx context.Context, allocationID string, paths [
 
 	// root reference_objects with parent_path=""
 	db = db.Or("parent_path = ? AND allocation_id = ?", "", allocationID)
-	err := db.Order("level, lookup_hash").Find(&refs).Error
+	err := db.Order("path, level").Find(&refs).Error
```
lpoli: Lets remove `level` as path is unique

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -18,8 +18,71 @@ type ReferencePath struct {
...
+
+func GetReferencePath2(ctx context.Context, allocationID, path string) (*Ref, error) {
+	return GetReferenceForHashCalculationFromPaths(ctx, allocationID, []string{path})
+}
+
+// GetReferenceForHashCalculationFromPaths validate and build full dir tree from db, and CalculateHash and return root Ref without saving in DB
+func GetReferenceForHashCalculationFromPaths(ctx context.Context, allocationID string, paths []string) (*Ref, error) {
+	var refs []Ref
+	db := datastore.GetStore().GetTransaction(ctx)
+	db = db.Select("id", "allocation_id", "type", "name", "path",
```
lpoli, line 29: Lets use gorm's smart select field. We can create a file for it where we can put smart select fields.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+	db := datastore.GetStore().GetTransaction(ctx)
+	db = db.Select("id", "name", "path", "hash", "size", "merkle_root")
+	err := db.Where(&Ref{AllocationID: allocationID, LookupHash: path_hash}).First(ref).Error
+	if err == nil {
+		return ref, nil
+	}
+	return nil, err
+}
+
+func GetReferenceForCopyFromLookupHash(ctx context.Context, allocationID, path_hash string) (*Ref, error) {
```
lpoli: This function is identical with `GetReferencePathFromLookupHash`. Can remove this function. We can add it later on if there are different parameters

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+	db := datastore.GetStore().GetTransaction(ctx)
+	db = db.Select("path", "name", "size", "hash", "merkle_root")
+	err := db.Where(&Ref{AllocationID: allocationID, Path: path}).First(ref).Error
+	if err == nil {
+		return ref, nil
+	}
+	return nil, err
+}
+
+func GetOnlyReferencePathFromLookupHash(ctx context.Context, allocationID, path_hash string) (*Ref, error) {
```
lpoli: `path_hash` should be `pathHash` as per go convention

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+	db = db.Select("id", "on_cloud", "size")
+	err := db.Where(&Ref{AllocationID: allocationID, Path: path}).First(ref).Error
+	if err == nil {
+		return ref, nil
+	}
+	return nil, err
+}
+
+// GetReferenceHash get FileRef ID with allcationID and path from postgres
+func GetReferenceHash(ctx context.Context, allocationID, path string) (*Ref, error) {
```
lpoli: If `hash` is only required then its better to return `hash` rather than `ref`. decoding to `ref` is costlier

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -18,8 +18,71 @@ type ReferencePath struct {
 	Ref  *Ref
 }
 
-func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, error) {
-	return GetReferencePathFromPaths(ctx, allocationID, []string{path})
+//func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, error) {
```
lpoli: Remove commented code

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -18,8 +18,71 @@ type ReferencePath struct {
...
+	for i := 1; i < len(refs); i++ {
+		if _, ok := refMap[refs[i].ParentPath]; !ok {
+			return nil, common.NewError("invalid_dir_tree", "DB has invalid tree.")
+		}
+		if _, ok := refMap[refs[i].Path]; !ok {
+			refMap[refs[i].ParentPath].AddChild(&refs[i])
+			refMap[refs[i].Path] = &refs[i]
+		}
+	}
+	if _, err := refs[0].CalculateHash(ctx, false); err != nil {
```
lpoli: if we are not storing it in db then do we even need to calculate hash? Its already there, right?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: @abouroumine This changes may be missed. Please check this

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: @abouroumine Did you consider this?

code/go/0chain.net/blobbercore/allocation/newfilechange.go
```diff
@@ -55,12 +55,12 @@ type NewFileChange struct {
 func (nf *NewFileChange) CreateDir(ctx context.Context, allocationID, dirName, allocationRoot string) (*reference.Ref, error) {
 	path := filepath.Clean(dirName)
 	tSubDirs := reference.GetSubDirsFromPath(path)
-
-	rootRef, err := reference.GetReferencePath(ctx, allocationID, nf.Path)
+	// Maybe Change this from GetReferencePath to GetReferencePath2
```
lpoli: Remove this comment

lpoli (changes requested):
I only see few changes to be made. Otherwise it looks good to me

code/go/0chain.net/blobbercore/handler/file_command_add.go
```diff
@@ -3,12 +3,12 @@ package handler
 import (
 	"context"
 	"encoding/json"
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/reference"
```
peterlimg: run goimports to format

code/go/0chain.net/blobbercore/handler/file_command_add.go
```diff
@@ -36,9 +38,11 @@ func (cmd *AddFileCommand) IsAuthorized(ctx context.Context, req *http.Request,
 		return common.NewError("invalid_parameters",
 			"Invalid parameters. Error parsing the meta data for upload."+err.Error())
 	}
-	exisitingFileRef, _ := reference.GetReference(ctx, allocationObj.ID, fileChanger.Path)
-
-	if exisitingFileRef != nil {
+	cmd.existingFileRef, err = reference.GetReferenceID(ctx, allocationObj.ID, fileChanger.Path)
```
peterlimg: What's the purpose for saving this existingFileRef to the cmd, if the file is duplicate, we will return error, and it won't be used.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -590,7 +587,8 @@ func (fsh *StorageHandler) RenameObject(ctx context.Context, r *http.Request) (i
 	mutex.Lock()
 	defer mutex.Unlock()
 
-	objectRef, err := reference.GetReferenceFromLookupHash(ctx, allocationID, pathHash)
+	// Changed From GetReferenceFromLookupHash To GetReferenceForHashCalculationFromPaths
```
peterlimg: Incorrect comment

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -947,8 +948,18 @@ func (fsh *StorageHandler) WriteFile(ctx context.Context, r *http.Request) (*blo
 	}
 
 	allocationID := allocationObj.ID
-	fileOperation := getFileOperation(r)
-	existingFileRef := getExistingFileRef(fsh, ctx, r, allocationObj, fileOperation)
+	cmd := createFileCommand(r)
+	err2 := cmd.IsAuthorized(ctx, r, allocationObj, clientID)
```
peterlimg: Check err2 right after it is returned to avoid unnecessary code checking, also, use `err` instead of `err2`

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -124,12 +124,8 @@ func (fsh *StorageHandler) GetAllocationUpdateTicket(ctx context.Context, r *htt
 	return allocationObj, nil
 }
 
-func (fsh *StorageHandler) checkIfFileAlreadyExists(ctx context.Context, allocationID, path string) *reference.Ref {
-	fileReference, err := reference.GetReference(ctx, allocationID, path)
-	if err != nil {
-		return nil
-	}
-	return fileReference
+func (fsh *StorageHandler) checkIfFileRefIDAlreadyExists(ctx context.Context, allocationID, path string) (*reference.Ref, error) {
```
peterlimg: No  need  to change function name

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+	db := datastore.GetStore().GetTransaction(ctx)
+	db = db.Select("hash")
+	err := db.Where(&Ref{AllocationID: allocationID, Path: path}).First(ref).Error
+	if err == nil {
+		return ref, nil
+	}
+	return nil, err
+}
+
+// GetReferenceDelete get FileRef ID with allcationID and path from postgres
```
peterlimg: Please comment appropriately

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+func GetReferenceDelete(ctx context.Context, allocationID, path string) (*Ref, error) {
```
peterlimg: What does the `Delete` mean in the function name, I don't see any delete relate action

code/go/0chain.net/blobbercore/reference/ref.go
peterlimg: GetReferencePathByLookupHash() would be better, also path_hash to pathHash

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,6 +224,109 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+	db := datastore.GetStore().GetTransaction(ctx)
+	db = db.Select("id", "path")
+	err := db.Where(&Ref{AllocationID: allocationID, LookupHash: path_hash}).First(ref).Error
+	if err == nil {
+		return ref, nil
+	}
+	return nil, err
+}
+
+func GetReferencePathFromLookupHash(ctx context.Context, allocationID, path_hash string) (*Ref, error) {
```
peterlimg: GetReferenceByLookupHash(), it does not only return path

code/go/0chain.net/blobbercore/reference/referencepath.go
```diff
@@ -18,8 +18,71 @@ type ReferencePath struct {
 	Ref  *Ref
 }
 
-func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, error) {
-	return GetReferencePathFromPaths(ctx, allocationID, []string{path})
+//func GetReferencePath(ctx context.Context, allocationID, path string) (*Ref, error) {
+//	return GetReferencePathFromPaths(ctx, allocationID, []string{path})
+//}
+
+func GetReferencePath2(ctx context.Context, allocationID, path string) (*Ref, error) {
```
peterlimg: Rename to GetReferencePath if it is used.

code/go/0chain.net/blobbercore/stats/filestats.go
```diff
@@ -2,7 +2,6 @@ package stats
 
 import (
 	"context"
-
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/datastore"
```
peterlimg, line 5: run  goimports to format

peterlimg (changes requested):

code/go/0chain.net/blobbercore/handler/file_command_add.go
abouroumine: to avoid redundant call to db for checking if file exists in both object_operation_handler => WriteFile function and file_command_add IsAuthorized.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
abouroumine: the reason I'm checking the err2 later is due to multiple system, unit and integration tests failures. since we are calling the IsAuthorized much ahead now to reduce db requests.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -72,7 +72,7 @@ func (store *postgresStore) GetDB() *gorm.DB {
 }
 
 func (store *postgresStore) AutoMigrate() error {
-
+	time.Sleep(20 * time.Second)
```
lpoli: Please remove this

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -783,12 +778,16 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
...
 
-	objectRef, err := reference.GetReferenceFromLookupHash(ctx, allocationID, pathHash)
+	objectRef, err := reference.GetLimitedRefFieldsByLookupHash(ctx, allocationID, pathHash, []string{"id", "name", "path", "hash", "size", "merkle_root"})
 
 	if err != nil {
 		return nil, common.NewError("invalid_parameters", "Invalid file path. "+err.Error())
 	}
 	newPath := filepath.Join(destPath, objectRef.Name)
+	destRef, _ := reference.GetLimitedRefFieldsByPath(ctx, allocationID, newPath, []string{"id"})
+	if destRef != nil {
```
lpoli: I think we should remove this because this is already checked in Line 807 below.

code/go/0chain.net/blobbercore/reference/object.go
```diff
@@ -17,9 +17,10 @@ func LoadObjectTree(ctx context.Context, allocationID, path string) (*Ref, error
 		GetTransaction(ctx)
 
 	path = filepath.Join("/", path)
-	db = db.Where("allocation_id = ? and deleted_at IS NULL and path LIKE ? ", allocationID, (path + "%"))
+	db = db.Where("allocation_id = ? and deleted_at IS NULL and path LIKE ? ", allocationID, path+"%")
 
-	db = db.Order("level desc, lookup_hash")
+	//db = db.Order("level desc, lookup_hash")
+	db = db.Order("level desc, path")
```
lpoli: Maybe `db.Order("path")` instead of `db.Order("level desc, path")` ?

lpoli (changes requested):
In one place you've used `db.Order("level desc, path")`. Please check for this.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
abouroumine: it did cause some system tests errors due to the order of the errors handling.

code/go/0chain.net/blobbercore/reference/object.go
abouroumine: db = db.Order("path desc") or db = db.Order("path") result into allocation root error for some system tests. the order by level desc is very important here during the deletion process.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | blobber/issue-524  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/reference/object.go
lpoli: Yes its because of below logic L#29-L#40.
It adds child in backward direction.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | blobber/issue-524  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

lpoli (approved):

code/go/0chain.net/blobbercore/allocation/renamefilechange.go
```diff
@@ -25,7 +25,12 @@ func (rf *RenameFileChange) DeleteTempFile() error {
 }
 
 func (rf *RenameFileChange) ApplyChange(ctx context.Context, change *AllocationChange, allocationRoot string) (*reference.Ref, error) {
-	isFilePresent, _ := reference.PathExists(ctx, rf.AllocationID, rf.NewName)
+
+	isFilePresent, err := reference.IsRefExist(ctx, rf.AllocationID, rf.NewName)
+	if err != nil {
+		Logger.Info("invalid_reference_path", zap.Any("error", err))
```
peterlimg: Log error in this way:
```go
Logger.Error("invalid_reference_path", zap.Error(err))
```

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -834,17 +833,19 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
...
+
 	path := r.FormValue("path")
 	if path == "" {
 		return nil, common.NewError("invalid_parameters", "Invalid path")
 	}
-
-	fileRef, _ := reference.GetReference(ctx, connectionObj.AllocationID, path)
+	fileRef, err := reference.GetLimitedRefFieldsByPath(ctx, connectionObj.AllocationID, path, []string{"path", "name", "size", "hash", "merkle_root"})
+	if err != nil {
+		Logger.Info("invalid_file", zap.Any("error", err))
```
peterlimg: Log error with Logger.Error(.., zap.Error(err))

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -897,11 +897,15 @@ func (fsh *StorageHandler) CreateDir(ctx context.Context, r *http.Request) (*blo
 		return nil, common.NewError("invalid_parameters", "Invalid dir path passed")
 	}
 
+	exisitingRef, err := fsh.checkIfFileAlreadyExists(ctx, allocationID, dirPath)
+	if err != nil {
+		Logger.Info("Error file reference", zap.Any("error", err))
```
peterlimg: Logger.Error(.., zap.Error(err))

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
peterlimg: If we got error from `IsAuthorized()` and do not check and return early, the execution of the following `IsACollaborator()`, `verifySignatureFromRequest()` would be a waste of time/resources, and moving `IsAuthorized()` ahead would not make sense. 

We should update unit tests and system tests instead of posting the the error check.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,36 +223,38 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
+func GetLimitedRefFieldsByLookupHash(ctx context.Context, allocationID, lookupHash string, selectedFields []string) (*Ref, error) {
+	ref := &Ref{}
 	db := datastore.GetStore().GetTransaction(ctx)
-	err := db.Select("type").Where("allocation_id=? AND path=?", allocationID, path).First(ref).Error
-	if err != nil {
-		return "", err
+	db = db.Select(selectedFields)
+	err := db.Where(&Ref{AllocationID: allocationID, LookupHash: lookupHash}).First(ref).Error
+	if err == nil {
+		return ref, nil
```
peterlimg: Please handle the error in the Go idiomatic way, i.e. check if `err != nil` to return error early

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -225,36 +223,38 @@ func GetReference(ctx context.Context, allocationID, path string) (*Ref, error)
...
-	ref := new(Ref)
+func GetReferenceByLookupHash(ctx context.Context, allocationID, pathHash string) (*Ref, error) {
+	ref := &Ref{}
 	db := datastore.GetStore().GetTransaction(ctx)
-	err := db.Select("id").Where("allocation_id=? AND path=?", allocationID, path).First(ref).Error
-	if err != nil {
-		return nil, err
+	err := db.Where(&Ref{AllocationID: allocationID, LookupHash: pathHash}).First(ref).Error
+	if err == nil {
+		return ref, nil
```
peterlimg: Check and return error early

code/go/0chain.net/blobbercore/allocation/attributesfilechange.go
```diff
@@ -38,13 +39,14 @@ func (ac *AttributesChange) ApplyChange(ctx context.Context, _ *AllocationChange
 		dirRef    = ref
 		treelevel = 0
 	)
-
+	dirRef.HashToBeComputed = true
```
peterlimg, line 42: What's rule for setting this `HashToBeComputed`, i.e when should we set it to true, and when to false

peterlimg (changes requested):

code/go/0chain.net/blobbercore/allocation/attributesfilechange.go
abouroumine: For the HashToBeComputed flag parameter is the result of this: https://github.com/0chain/blobber/pull/529#discussion_r816455077 
since whenever there's new/update/delete file we have to update the hash but we can use hashes that did not change to optimise the calculations.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | blobber/issue-524  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | blobber/issue-524  |
| gosdk  | blobber/issue-524  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | blobber/issue-524  |
| gosdk  | blobber/issue-524  |
| zboxcli  | blobber/issue-524  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | issue/524  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/handler/handler_test.go
```diff
@@ -248,6 +235,15 @@ func isEndpointRequireSignature(name string) bool {
 	}
 }
 
+func isEndpointUpload(name string) bool {
+	switch name {
+	case "Upload":
+		return true
+	default:
+		return false
+	}
```
peterlimg: Could be a simple `return if name == "Upload"` unless there are other possible names can be treated as Upload.

code/go/0chain.net/blobbercore/handler/handler_test.go
```diff
@@ -35,6 +24,15 @@ import (
...
 	"github.com/0chain/blobber/code/go/0chain.net/core/logging"
+	"github.com/0chain/gosdk/core/zcncrypto"
+	"github.com/0chain/gosdk/zboxcore/client"
+	"github.com/0chain/gosdk/zboxcore/fileref"
+	"github.com/0chain/gosdk/zboxcore/marker"
+	"github.com/0chain/gosdk/zcncore"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
```
peterlimg, line 26: What format tools are you using, we'd better run `goimports` so that all have the same code format

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
peterlimg: Please fix this

peterlimg (changes requested):

code/go/0chain.net/blobbercore/handler/handler_test.go
abouroumine: Goland IDE => Code Style => Go => goimports.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -809,17 +806,19 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
...
+
 	path := r.FormValue("path")
 	if path == "" {
 		return nil, common.NewError("invalid_parameters", "Invalid path")
 	}
-
-	fileRef, _ := reference.GetReference(ctx, connectionObj.AllocationID, path)
+	fileRef, err := reference.GetLimitedRefFieldsByPath(ctx, connectionObj.AllocationID, path, []string{"path", "name", "size", "hash", "merkle_root"})
+	if err != nil {
+		Logger.Info("invalid_file", zap.Error(err))
```
peterlimg: Please log error with `Logger.Error()` as it will give stack info, so as other places.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -952,8 +953,22 @@ func (fsh *StorageHandler) WriteFile(ctx context.Context, r *http.Request) (*blo
...
+
+	var existingFileRef *reference.Ref
+	switch rCmd := cmd.(type) {
+	case *UploadFileCommand:
+		existingFileRef = nil
+	case *UpdateFileCommand:
+		existingFileRef = rCmd.existingFileRef
+	case *FileCommandDelete:
+		existingFileRef = rCmd.existingFileRef
+	}
```
peterlimg, line 971: The `*UpdateFileCommand` and `*FileCommandDelete` cases can be handled in one case in the following way:

```go
switch rCmd := cmd.(type) {
...
case *UpdateFileCommand, 
        *FileCommandDelete:
    existingFileRef = rcmd.existingFileRef
...
}
```

Also, please return error if the type was unknown.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
abouroumine: I think the code if I update it I should do something like this:
```
        case *UpdateFileCommand:
		existingFileRef = rCmd.existingFileRef
	case *FileCommandDelete:
		existingFileRef = rCmd.existingFileRef
	default:
		existingFileRef = nil
```

- We can't get the `rCmd.existingFileRef` if we don't specify exactly the type createFileCommand of the `rCmd`.
If we do something like this: 
```
 case *UpdateFileCommand, 
        *FileCommandDelete:
    existingFileRef = rcmd.existingFileRef
```
the rcmd.existingFileRef is not recognized.
- The Error is handled before we reach this step in:
```
        err = cmd.IsValidated(ctx, r, allocationObj, clientID)
	if err != nil {
		return nil, err
	}
	var existingFileRef *reference.Ref
	switch rCmd := cmd.(type) {
	case *UpdateFileCommand:
		existingFileRef = rCmd.existingFileRef
	case *FileCommandDelete:
		existingFileRef = rCmd.existingFileRef
	default:
		existingFileRef = nil
	}
```
Here we just affect to the existingFileRef for Checking if possible Collaborator.

peterlimg (approved):
lgtm

moldis (approved):

service-0chain:
Manual system tests [cancelled] with the following config
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #539, opened 2022-02-15
Feat/db auto migrate

### LINKED ISSUE #230: [FEATURE] use database migrate tool instead of pure sql scripts

cnlangzi:
- [x]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [x]  Merge your system tests PR to master BEFORE merging this PR

- zbox:
- zwalelet:
- gosdk:
- 0chain:
- system_test:

### DIFF
--- a/code/go/0chain.net/blobber/http.go
+++ b/code/go/0chain.net/blobber/http.go
@@ -38,7 +38,7 @@ func startHttpServer() {
 	go startServer(&wg, r, mode, httpsPort, true)
 
 	logging.Logger.Info("Ready to listen to the requests")
-	fmt.Println("[11/11] start http server	[OK]")
+	fmt.Println("[12/12] start http server	[OK]")
 
 	wg.Wait()
 }
--- a/code/go/0chain.net/blobbercore/datastore/postgres.go
+++ b/code/go/0chain.net/blobbercore/datastore/postgres.go
@@ -2,12 +2,14 @@ package datastore
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
 	"github.com/0chain/blobber/code/go/0chain.net/core/common"
-	. "github.com/0chain/blobber/code/go/0chain.net/core/logging"
+	"github.com/0chain/blobber/code/go/0chain.net/core/logging"
+	"go.uber.org/zap"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
 )
@@ -59,10 +61,51 @@ func (store *postgresStore) GetTransaction(ctx context.Context) *gorm.DB {
 	if conn != nil {
 		return conn.(*gorm.DB)
 	}
-	Logger.Error("No connection in the context.")
+	logging.Logger.Error("No connection in the context.")
 	return nil
 }
 
 func (store *postgresStore) GetDB() *gorm.DB {
 	return store.db
+}
+
+func (store *postgresStore) AutoMigrate() error {
+
+	err := store.db.AutoMigrate(&Migration{})
+	if err != nil {
+		logging.Logger.Error("[db]", zap.Error(err))
+	}
+
+	latest := &Migration{}
+	result := store.db.Raw(`select * from "migrations" order by "version" desc limit 1`).First(latest)
+
+	if result.Error != nil {
+		if errors.Is(gorm.ErrRecordNotFound, result.Error) {
+			latest.Version = "0.0.0"
+			latest.CreatedAt = time.Date(2021, 10, 14, 0, 0, 0, 0, time.UTC)
+			err = store.db.Create(latest).Error
+
+			if err != nil {
+				logging.Logger.Error("[db]"+latest.Version, zap.Error(err))
+				return err
+			}
+		} else {
+			logging.Logger.Error("[db]", zap.Error(result.Error))
+			return err
+		}
+	}
+
+	for i := 0; i < len(releases); i++ {
+		v := releases[i]
+		if v.After(latest) {
+			err = v.Migrate(store.db)
+			if err != nil {
+				logging.Logger.Error("[db]"+v.Version, zap.Error(err))
+				return err
+			} else {
+				logging.Logger.Info("[db]" + v.Version + " migrated")
+			}
+		}
+	}
+	return nil
 }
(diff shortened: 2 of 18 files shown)


code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -59,10 +64,51 @@ func (store *postgresStore) GetTransaction(ctx context.Context) *gorm.DB {
...
+
+func (store *postgresStore) AutoMigrate() error {
+
+	err := store.db.AutoMigrate(&Migration{})
+	if err != nil {
+		logging.Logger.Error("[db]", zap.Error(err))
+	}
+
+	latest := &Migration{}
+	result := store.db.Raw(`select * from "migrations" order by "version" desc limit 1`).First(latest)
```
ssardana96: version is a string
let's say a comparison of "21.1.0" and "3.1.0" happens. it'll consider "3.1.0" bigger among them

code/go/0chain.net/blobbercore/datastore/postgres_migration.go
```diff
@@ -0,0 +1,60 @@
...
+	Scripts   []string  `gorm:"-"`
+}
+
+// After check if migrate is newer than latest
+func (m *Migration) After(latest *Migration) bool {
+	currentVersions := strings.Split(m.Version, ".")
+	latestVersions := strings.Split(latest.Version, ".")
+
+	for i := 0; i < 3; i++ {
+		if currentVersions[i] > latestVersions[i] {
```
ssardana96: string comparisons are a bit tricker.
probably you'll have to check int comparison of these

ssardana96: what if currentVersion is 2.13.0 and latestVersion is 3.12.0
this will return true. but should it be true?

ssardana96 (dismissed):

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: it is not designed as normal version
- TABLE version when you add/drop any table,
- INDEX version when you add/drop/update and index
- COLUMN version when you add/drop/update any column
so any part is bigger, and it should be migrated

code/go/0chain.net/blobbercore/datastore/postgres_migration.go
cnlangzi: yes. because index is changed on this version. actually , it is 3 versions.
1st for table, 2nd for index, 3rd for columns.

code/go/0chain.net/blobbercore/datastore/postgres.go
ssardana96: no but in string
"3" > "21"
but ideally the logic with which you want to compare that is integer based, where 21 > 3

cnlangzi: hmm. got it. let me convert it to int first.

ssardana96: the above is not an issue because of versioning mechanism, but string sorting is different that integer sorting
which will fetch wrong latest version for you

cnlangzi: done

cnlangzi: get latest version by created_at instead of version

ssardana96:
looks good to me logically.
writing tests might help us validate these.

cnlangzi:
it is not easy to mock such query in postgres. We always mock query result directly.  I will check it by hand.

cnlangzi:
migration is rewritten by checking version if exists instead of version comparison. it is an easier and more stable solution.

code/go/0chain.net/blobber/http.go
```diff
@@ -38,7 +38,7 @@ func startHttpServer() {
 	go startServer(&wg, r, mode, httpsPort, true)
 
 	logging.Logger.Info("Ready to listen to the requests")
-	fmt.Println("[11/11] start http server	[OK]")
+	fmt.Println("[12/12] start http server	[OK]")
```
princeparmar, line 41: Can we not have one common method which task msg string as param and handle [current-step/total-step]

code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -59,10 +63,63 @@ func (store *postgresStore) GetTransaction(ctx context.Context) *gorm.DB {
...
+		} else {
+			err = v.Migrate(store.db)
+			if err != nil {
+				logging.Logger.Error("[db]"+v.Version, zap.Error(err))
+				return err
+			} else {
+				logging.Logger.Info("[db]" + v.Version + " migrated")
+				fmt.Print("	[OK]\n")
+			}
+		}
```
princeparmar, line 96: if hasMigrated {
			fmt.Print("	[SKIP]\n")
                        continue
		} 
		err = v.Migrate(store.db)
		if err != nil {
			logging.Logger.Error("[db]"+v.Version, zap.Error(err))
			return err
		} else {
			logging.Logger.Info("[db]" + v.Version + " migrated")
			fmt.Print("	[OK]\n")
		}

code/go/0chain.net/blobber/http.go
cnlangzi: It seems be not very worth. because current-step always be changed. but total-setp can be a variables.

cnlangzi: I am not sure if it is worth to spend time to extract it to method for `totalSetps`. Because it is not a functional method.  I would suggest to leave it as it is.

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: do you mean add `continue` ?  BTW you can use suggestion  block. try it next time.

cnlangzi: improved loop code

princeparmar (approved):

Issue #507, opened 2022-01-17
Race condition with read marker


lpoli:
Current ReadMarker structure:
```
type ReadMarker struct {
	ClientID        string           `json:"client_id"`
	ClientPublicKey string           `json:"client_public_key"`
	BlobberID       string           `json:"blobber_id"`
	AllocationID    string           `json:"allocation_id"`
	OwnerID         string           `json:"owner_id"`
	Timestamp       common.Timestamp `json:"timestamp"`
	ReadCounter     int64            `json:"counter"`
	Signature       string           `json:"signature"`
}
```

As per the protocol _**The smart contract will release funds corresponding to the difference between the read_counter in the marker and the latest read_counter redeemed on the blockchain**_

The importance of ReadCounter is only to allow blobber to redeem tokens from read pools. The ReadCounter is based on the assumption that each counter corresponds to 64KB read by the client. So the price calculation is done as: `ΔReadcounter * blockSize * read_price`. 

_Note: BlockSize is customizable with new chunkupload feature so it is no more fixed 64KB._

However above implementation leads to race condition for ReadCounter as readmarker signed by client requires it to be the latest ReadCounter so that blobber can calculate price easily. Blobber code rejects download request if readCounter is not in sync.
```
	if latestRM != nil &&
		latestRM.ReadCounter+(numBlocks) != readMarker.ReadCounter {

		var response = &blobberhttp.DownloadResponse{
			Success:      false,
			LatestRM:     latestRM,
			Path:         fileref.Path,
			AllocationID: fileref.AllocationID,
		}
		return response, nil
	}
```

ReadMarker is associated to client so there won't be race between clients for ReadCounter as they will have their own ReadCounter. However if a client sends multiple download requests, be it for same file or different files, there will be race for latest ReadCounter between requests.

In my view this problem can be solved by replacing ReadCounter with ReadSize and requiring blobber to request read pools with each read marker to redeem read tokens rather than single read marker being able to redeem whole tokens. With this client simply signs readmarker with required size and there is no race for anything.

EDIT:
After successfully merging the above solution, we realized that there is no mechanism against indefinite redeeming of same read marker. `ReadCounter` had served this purpose. So we moved

lpoli:
cc @taustin @guruhubb @Sriep @peterlimg  @moldis @cnlangzi

Sriep:
I suggest doing this in three stages.
1. Blobber: Add the `read_size` field to the ReadMarker.
[...]

PR #552, opened 2022-02-23
Modify/read marker

lpoli:
This PR contains the following changes:
1. Change ReadMarker's ReadCounter to ReadSize
2. Remove latest ReadMarker logic from the code as it is useless with current implementation
3. Modify `pendings` table to accommodate above changes
4. Reduce unnecessary functions and function calls in worker in `readmarker` package

This PR solves the issue https://github.com/0chain/blobber/issues/507


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | modify/read-marker  |
| zwalletcli  | staging  |
| 0chain  | modify/read-marker  |
| blobber  | modify/read-marker  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -49,129 +49,122 @@ func readPreRedeem(ctx context.Context, alloc *allocation.Allocation, numBlocks,
...
+		until     = common.Now() + common.Timestamp(config.Configuration.ReadLockTimeout)
 
-		want = alloc.WantRead(blobberID, numBlocks)
+		currentReadSize = numBlocks * allocation.CHUNK_SIZE // currently only default size is used as chunk size
 
 		rps []*allocation.ReadPool
 	)
 
-	if want == 0 {
+	if currentReadSize == 0 || !(alloc.GetRequiredReadBalance(blobberID, currentReadSize) > 0) {
```
peterlimg: It would be easier to read with alloc.GetRequiredReadBalance(...) <= 0

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -101,102 +107,105 @@ func (a *Allocation) WantWrite(blobberID string, size int64, wmt common.Timestam
...
+	defer lock.Unlock()
+
+	pending := new(Pending)
+	err = db.Model(&Pending{}).Where("id=?", key).First(pending).Error
+	if err == nil && pending != nil {
+		pending.PendingWrite += pendingWrite
+		pending.PendingRead += pendingRead
+	} else {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			pending.ID = clientID + ":" + allocationID
```
peterlimg: use the defined `key` above

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -101,102 +107,105 @@ func (a *Allocation) WantWrite(blobberID string, size int64, wmt common.Timestam
...
+	key := clientID + ":" + allocationID
+	// Lock is required because two process can simultaneously call this function and read pending data
+	// thus giving same value leading to inconsistent data
+	lock := pendingMapLock.GetLock(key)
+	lock.Lock()
+	defer lock.Unlock()
+
+	pending := new(Pending)
+	err = db.Model(&Pending{}).Where("id=?", key).First(pending).Error
+	if err == nil && pending != nil {
```
peterlimg: is it possible err == nil and pending == nil, if so, the current logic would return nil but no data will be saved.

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -101,102 +107,105 @@ func (a *Allocation) WantWrite(blobberID string, size int64, wmt common.Timestam
...
+func AddToPending(db *gorm.DB, clientID, allocationID string, pendingWrite, pendingRead int64) (err error) {
+	key := clientID + ":" + allocationID
+	// Lock is required because two process can simultaneously call this function and read pending data
+	// thus giving same value leading to inconsistent data
+	lock := pendingMapLock.GetLock(key)
+	lock.Lock()
+	defer lock.Unlock()
+
+	pending := new(Pending)
+	err = db.Model(&Pending{}).Where("id=?", key).First(pending).Error
```
peterlimg, line 179: Perhaps use swithc case can make the error handling easier to read:

switch err {
case nil:
    // update existing
case gormErrRecordNotFound:
    // add new
default:
    return err
}

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -49,129 +49,122 @@ func readPreRedeem(ctx context.Context, alloc *allocation.Allocation, numBlocks,
...
+		requiredBalance = alloc.GetRequiredWriteBalance(blobberID, writeMarker.Size, writeMarker.Timestamp)
 
-		pend *allocation.Pending
-		wps  []*allocation.WritePool
+		wps []*allocation.WritePool
 	)
 
-	if writeMarker.Size <= 0 || want <= 0 {
-		return // skip if write price is zero or it's about deleting
+	if writeMarker.Size <= 0 || !(requiredBalance > 0) {
```
peterlimg: unnecessary !, could be requiredBalance <= 0

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -49,129 +49,122 @@ func readPreRedeem(ctx context.Context, alloc *allocation.Allocation, numBlocks,
...
+		return
 	}
 
-	pend, err = allocation.GetPending(db, payerID,
-		alloc.ID, blobberID)
+	writePoolBalance, err := allocation.GetWritePoolsBalance(db, writeMarker.ClientID, alloc.ID, until)
```
peterlimg: should be writeMarker.ClientID or payerID?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -49,129 +49,122 @@ func readPreRedeem(ctx context.Context, alloc *allocation.Allocation, numBlocks,
...
 	if err != nil {
-		return common.NewErrorf("write_pre_redeem",
-			"can't get pending payments: %v", err)
+		Logger.Error(err.Error())
+		return common.NewError("write_pre_redeem", "database error while getting write pool balance")
 	}
 
-	wps, err = pend.WritePools(db, blobberID, until)
+	pendingWriteSize, err := allocation.GetPendingWrite(db, writeMarker.ClientID, alloc.ID)
```
peterlimg: writeMarker.ClientID or payerID?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -49,129 +49,122 @@ func readPreRedeem(ctx context.Context, alloc *allocation.Allocation, numBlocks,
...
+			alloc.ID, writeMarker.BlobberID, writePoolBalance, requiredBalance)
 	}
 
-	// update pending writes: add size to redeem to (not tokens)
-	pend.AddPendingWrite(writeMarker.Size)
-	if err = pend.Save(db); err != nil {
-		return common.NewErrorf("write_pre_redeem",
-			"can't save pending writes in DB: %v", err)
-	}
+	if err := allocation.AddToPending(db, writeMarker.ClientID, alloc.ID, writeMarker.Size, 0); err != nil {
```
peterlimg: writeMarke.ClientID or payerID ?

code/go/0chain.net/blobbercore/readmarker/authticket.go
```diff
@@ -0,0 +1,69 @@
...
+	RefType         string           `json:"reference_type"`
+	Expiration      common.Timestamp `json:"expiration"`
+	Timestamp       common.Timestamp `json:"timestamp"`
+	ReEncryptionKey string           `json:"re_encryption_key"`
+	Signature       string           `json:"signature"`
+	Encrypted       bool             `json:"encrypted"`
+}
+
+func (rm *AuthTicket) GetHashData() string {
+	hashData := fmt.Sprintf("%v:%v:%v:%v:%v:%v:%v:%v:%v:%v:%v", rm.AllocationID, rm.ClientID, rm.OwnerID, rm.FilePathHash, rm.FileName, rm.RefType, rm.ReEncryptionKey, rm.Expiration, rm.Timestamp, rm.ActualFileHash, rm.Encrypted)
```
peterlimg: Please do not put all in one line, format with multiple lines appropriately. check https://wiki.crdb.io/wiki/spaces/CRDB/pages/181371303/Go+Golang+coding+guidelines#Line-Length

code/go/0chain.net/blobbercore/readmarker/worker.go
```diff
@@ -3,74 +3,117 @@ package readmarker
...
 
-	Logger.Info("successfully redeemed read marker",
-		zap.Any("rm", rmEntity.LatestRM))
-	return
-}
+	guideCh := make(chan struct{}, config.Configuration.RMRedeemNumWorkers)
+	wg := sync.WaitGroup{}
+	for _, rme := range rms {
+		rme.ReadMarker.BlobberID = node.Self.ID
+		guideCh <- struct{}{}
```
peterlimg, line 83: Seems the guideCh is not needed

code/go/0chain.net/blobbercore/readmarker/worker.go
```diff
@@ -3,74 +3,117 @@ package readmarker
...
+		wg.Add(1)
+		go func(rme *ReadMarkerEntity) {
+			defer func() {
+				<-guideCh
+				wg.Done()
+			}()
+			ctx, ctxCncl := context.WithCancel(rctx)
+			defer ctxCncl()
```
peterlimg, line 91: CancelContext seems not necessary here, I don't see code that require cancel or handle the cancel error, so please use context.Background() instead.

code/go/0chain.net/blobbercore/readmarker/worker.go
```diff
@@ -79,7 +122,13 @@ func startRedeemMarkers(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			redeemReadMarker(ctx)
+			newCtx, ctxCncl := context.WithCancel(ctx)
```
peterlimg, line 125: Unnecessary context.WithCancel

code/go/0chain.net/core/common/lock.go
```diff
@@ -0,0 +1,78 @@
+package common
+
+import (
+	"sync"
+)
+
+type Lock struct {
```
peterlimg, line 7: If this lock is still in experiment stage, and not used for this PR, we'd better do it in a separate PR.

peterlimg (changes requested):
Any reason why do we remove the blobber id from the `pendings` table, or is there anything that I missed that this had been discussed?

lpoli:
> Any reason why do we remove the blobber id from the `pendings` table, or is there anything that I missed that this had been discussed?

We can always get blobber id with `node.Self.Id`. It would simply occupy disk space and is always same. That's why I removed along with @sculptex suggestion on db optimizations.

code/go/0chain.net/core/common/lock.go
lpoli: I've used it in `AddPending` function. If you suggest to use cache version then I will change this.
What should I do?

code/go/0chain.net/blobbercore/allocation/entity.go
lpoli: Actually in line 179 pending is already not nil.
I will fix this as if pending is to be nil then err is guaranteed

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: I had thought about it while writing but don't remember now why I did that.
I will change as per your suggestion

code/go/0chain.net/blobbercore/readmarker/worker.go
lpoli: GuideCh is buffered channel of `config.Configuration.RMRedeemNumWorkers` so it won't let more than this number of goroutines to run at a time. There can be huge number of readMarkers to redeem at a time

peterlimg: ah okay, got it.

code/go/0chain.net/core/common/lock.go
```diff
@@ -0,0 +1,78 @@
...
+	// Parent Map
+	pMap *MapLocker
+}
+
+// Lock Acquire lock
+func (l *Lock) Lock() {
+	for {
+		l.actualLock.Lock()
+		if l.stale {
+			l = l.pMap.GetLock(l.key)
```
peterlimg: This lock assignment of `l` will not work, it tries to change the receiver pointer. It does not work, just like we try to change the pointer of a passed in parameter of a function. So the result would be we still using the staled lock even we actually get the new one from pMap, while this staled lock was actually had been removed from pMap when the `Unlock()` was called, so using this stale lock is dangerous as it does not guarantee the thread safe of resources that it suppose to protect under the same key.

code/go/0chain.net/core/common/lock.go
```diff
@@ -0,0 +1,78 @@
...
+	mu *sync.Mutex
+}
+
+func (m *MapLocker) GetLock(key string) (l *Lock) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	valueI, ok := m.m.Load(key)
+	if ok {
+		l = valueI.(*Lock)
+		l.countMu.Lock()
```
peterlimg, line 55: Instead of using mutex to protect the l.count, we can use sync/atomic to avoid potential deadlock

code/go/0chain.net/core/common/lock.go
```diff
@@ -0,0 +1,78 @@
...
+type MapLocker struct {
+	m *sync.Map
+	// Reduce updating lock with GetLock call with same key
+	mu *sync.Mutex
+}
+
+func (m *MapLocker) GetLock(key string) (l *Lock) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	valueI, ok := m.m.Load(key)
```
peterlimg, line 52: `lock.Unlock()` deletes the key from the `MapLocker` without the protect of the `m.mu` lock, imagine the situation that we acquired the `m.mu.Lock()` and load the lock from the map here, then in another goroutine, this lock was unlocked, and it's count reduce to 0, then it would be marked as `stale` and removed from the `MapLocker`. The goroutine comes back here again and continue, the lock's `count` updated, and returned to user. However, it was not in the MapLocker anymore, we could have thread safe issue if user try to get the lock of the same key again in a new goroutine, as the MapLocker will give a new one, thus we will try to lock the same resources with different locks, which obviously, would not work as expected.

peterlimg (changes requested):

code/go/0chain.net/core/common/lock.go
lpoli: >The goroutine comes back here again and continue, the lock's `count` updated, and returned to user. However, it was not in the MapLocker anymore,

@peterlimg I don't see how lock's `count` will get updated if it does not exist in the `mapLocker`. Only one goroutine will have stale lock but it will check against staleness and calls `GetLock` again. It will get new lock with `count=1`. `GetLock` is protected by `m.mu.Lock` and further `l.countMu.Lock` will protect count update.

code/go/0chain.net/core/common/lock.go
lpoli: This is severe issue. Thanks a lot for pointing it out.
If I can fix this issue, I suppose I can use this locker. Do you suggest so?

code/go/0chain.net/blobbercore/readmarker/worker.go
lpoli: Use of context to get and create transaction also seems unnecessary. I simply used `rctx` as a parent so even if later on we make good use of context then having `rctx` as a parent might be useful.

code/go/0chain.net/core/common/lock.go
peterlimg: The lock was acquired by another thread. Let me explain in details and see the case below:

1. A gorountine called the `GetLock()` and get a lock `A`, it will use `A.Lock()` then `A.Unlock()` to release the lock. So suppose we called `A.Lock()` but hadn't called the `A.Unlock()`
2. Another gorountine come to the `mapLocker.GetLock()` function as show above, this goroutine will get the lock `A`  as well if the same key was used. This goroutine will get the lock `A` and suppose it come  to the code in line `52` where it just loaded the `valueI` from the map, then it go back to the previous goroutine.
3. The previosus goroutine called the `A.Unlock()` as show below, apparently, the `count` would be reduced to `0` and  then it will delete it self from the map locker:
    ```go
    func (l *Lock) Unlock() {
	        l.actualLock.Unlock()
	        l.countMu.Lock()
	        l.count--
	        if l.count == 0 {
		        l.stale = true
		        l.pMap.m.Delete(l.key)
	        }
	        l.countMu.Unlock()
    }
    ```
4. Come back to the newer goroutine, it continue as the code show above, this lock `A` will be returned to the user, but it does not exist in the map anymore.

lpoli: @peterlimg While unlocking, a goroutine also achieves `l.countMu.Lock()` and reduce count by 1. In `GetLock` function it should also get `l.countMu.Lock` to update count.

Suppose `A.Unlock()` reduce count to 0, made itself `stale` and removed it from `mapLocker`. When another goroutine which had got same lock when tries to achieve `A.Lock` it also checks if it is `stale` in line 24. This `staleness` should ensure that there are no two locks for same key

peterlimg: @lpoli yes, you're right, the stale checking should be able to ensure no two locks for the same key. So this looks good to me as long as the previous `locker pointer` assignment issue is fixed.

PR #571, opened 2022-03-03
Fix file upload inside file issue

lpoli:
This PR fixes issue of file upload inside a file along with directory creation inside a file path

PR #575, opened 2022-03-07
fix(upload): validate path

cnlangzi:
### Fixes
- it will be rejected if a file is uploaded inside a file

### Tests

### DIFF
--- a/code/go/0chain.net/blobbercore/reference/ref_walker.go
+++ b/code/go/0chain.net/blobbercore/reference/ref_walker.go
@@ -1,0 +1,146 @@
+package reference
+
+import (
+	"path"
+	"strings"
+)
+
+func NewRefWalkerFromPath(p string) *RefWalker {
+	return NewRefWalker(strings.Split(path.Clean(p), "/"))
+}
+
+// NewRefWalker wrap dirs as RefWaller
+func NewRefWalker(dirs []string) *RefWalker {
+
+	return &RefWalker{
+		items:  dirs,
+		length: len(dirs),
+		index:  0,
+	}
+}
+
+type RefWalker struct {
+	items  []string
+	length int
+	index  int
+}
+
+// Name get current dir name
+func (d *RefWalker) Name() string {
+	if d == nil {
+		return "/"
+	}
+
+	// current is root
+	if d.index == 0 {
+		return "/"
+	}
+
+	if d.index < d.length {
+		return d.items[d.index]
+	}
+
+	return "/"
+}
+
+// Path get current dir path
+func (d *RefWalker) Path() string {
+	if d == nil {
+		return ""
+	}
+
+	// current is root
+	if d.index == 0 {
+		return "/"
+	}
+
+	if d.index < d.length {
+		return "/" + path.Join(d.items[:d.index+1]...)
+	}
+
+	return ""
+}
+
+// Parent get curerent parent path
+func (d *RefWalker) Parent() string {
+	if d == nil {
+		return ""
+	}
+
+	// current is root
+	if d.index == 0 {
+		return ""
+	}
+
+	return "/" + path.Join(d.items[:d.index]...)
+}
+
+// Level get current dir level
+func (d *RefWalker) Level() int {
+	if d == nil {
+		return 0
+	}
+
+	return d.index + 1
+}
+
+// Top move to root dir
+func (d *RefWalker) Top() bool {
+	if d == nil {
+		return false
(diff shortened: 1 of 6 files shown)


lpoli:
@cnlangzi Actually this won't work. First of all, `util.SplitPath` does not return path's values for parent path but for file path i.e. for `/a/b/c/d/e.txt` it returns [a b c d e.txt] instead of [a b c d].

The other important missed point is, if there is already a file with path `/a/b/c.txt` and user uploaded a file with path `/a/b/c.txt/d/e/f/g.txt` then we cannot only check against parent path of the filepath. So we need to check for `/a`, `/a/b`, `/a/b/c.txt`, ....

I had considered this as well. If you make this changes then I will drop my PR.

cnlangzi:
> @cnlangzi Actually this won't work. First of all, `util.SplitPath` does not return path's values for parent path but for file path i.e. for `/a/b/c/d/e.txt` it returns [a b c d e.txt] instead of [a b c d].
> 
> The other important missed point is, if there is already a file with path `/a/b/c.txt` and user uploaded a file with path `/a/b/c.txt/d/e/f/g.txt` then we cannot only check against parent path of the filepath. So we need to check for `/a`, `/a/b`, `/a/b/c.txt`, ....
> 
> I had considered this as well. If you make this changes then I will drop my PR.

fixed both of them.

code/go/0chain.net/blobbercore/reference/ref_walker.go
```diff
@@ -0,0 +1,146 @@
...
+	if d.index < d.length {
+		return d.items[d.index]
+	}
+
+	return "/"
+}
+
+// Path get current dir path
+func (d *RefWalker) Path() string {
+	if d == nil {
```
peterlimg, line 48: Perhaps we should return error when d is nil, or when index is >= length, otherwise users will have to check the code to know what happened when the returned value is "".

code/go/0chain.net/blobbercore/reference/ref_walker.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if d.index < d.length {
+		return "/" + path.Join(d.items[:d.index+1]...)
+	}
+
+	return ""
+}
+
+// Parent get curerent parent path
+func (d *RefWalker) Parent() string {
```
peterlimg: Same as Path(), we'd better return errors on "" strings

code/go/0chain.net/blobbercore/util/strings_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			name:  "only 1 directory level deep",
+			path:  "/file1",
+			items: []string{"file1"},
+		},
+		{
+			name:  "nested directories",
+			path:  "/dir1/dir2/dir3/file1",
+			items: []string{"dir1", "dir2", "dir3", "file1"},
+		},
```
peterlimg, line 30: Adding cases for empty subpath perhaps, e.g. /dir1/dir2//dir3///dir4

peterlimg (changes requested):

code/go/0chain.net/blobbercore/reference/ref_walker.go
cnlangzi: fixed

cnlangzi: fixed

code/go/0chain.net/blobbercore/reference/ref_walker.go
cnlangzi: fixed

cnlangzi: fixed

code/go/0chain.net/blobbercore/util/strings_test.go
cnlangzi: fixed

cnlangzi:
@peterlimg  let's close this PR. and @lpoli will fix this bug on his PR https://github.com/0chain/blobber/pull/571. please go to review his PR instead of it. thanks.

cnlangzi:
@peterlimg  let's close this PR. and @lpoli will fix this bug on his PR https://github.com/0chain/blobber/pull/571. please go to review his PR instead of it. thanks.

cnlangzi:
sorry. it is https://github.com/0chain/blobber/pull/578. #571 seems is outdated.

PR #578, opened 2022-03-08
Fix file upload inside file

### LINKED ISSUE #454: File upload inside a file
When we upload a file to some remotepath it creates parent directories if they already are not present. So when a file is uploaded say `/a.txt` blobber stores it as file type `f`, which is absolutely fine.

Now when we upload a file say `b.txt` with remotepath `/a.txt/b.txt` then blobber happily accepts it and when we do `zbox list --remotepath /a.txt` we can see `b.txt` in its list. And when we do `zbox list --remotepath /`, it shows `a.txt` with type `f`.

Blobber can provide this feature/issue, but it results in inconsistencies and lots of confusion for other tools and direct user as well.
We should not allow upload inside a file.

lpoli:
This PR fixes issue of file upload inside a file along with directory creation inside a file path

### DIFF
--- a/code/go/0chain.net/blobbercore/handler/file_command_add.go
+++ b/code/go/0chain.net/blobbercore/handler/file_command_add.go
@@ -40,10 +40,9 @@ func (cmd *AddFileCommand) IsAuthorized(ctx context.Context, req *http.Request, 
 
 	if exisitingFileRef != nil {
 		return common.NewError("duplicate_file", "File at path already exists")
+	} else if err := validateParentPathType(ctx, allocationObj.ID, fileChanger.Path); err != nil {
+		return err
 	}
-	// else if err := validateParentPathType(ctx, allocationObj.ID, fileChanger.Path); err != nil {
-	// 	return err
-	// }
 
 	//create a FixedMerkleTree instance first, it will be reloaded from db in cmd.reloadChange if it is not first chunk
 	//cmd.fileChanger.FixedMerkleTree = &util.FixedMerkleTree{}
--- a/code/go/0chain.net/blobbercore/handler/file_command_add.go
+++ b/code/go/0chain.net/blobbercore/handler/file_command_add.go
@@ -3,7 +3,9 @@ package handler
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
+	"path/filepath"
 
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/allocation"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/blobberhttp"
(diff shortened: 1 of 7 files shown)


code/go/0chain.net/blobbercore/handler/file_command_add.go
```diff
@@ -40,6 +40,8 @@ func (cmd *AddFileCommand) IsAuthorized(ctx context.Context, req *http.Request,
 
 	if exisitingFileRef != nil {
 		return common.NewError("duplicate_file", "File at path already exists")
+	} else if err := validateParentPathType(ctx, allocationObj.ID, fileChanger.Path); err != nil {
```
peterlimg: We prefer do this in a new `if ` section instead of `if  .. else if..`.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -235,6 +235,59 @@ func GetReferenceFromLookupHash(ctx context.Context, allocationID, path_hash str
...
+		return nil, err
+	}
+	return ref, nil
+}
+
+// IsRefExist checks if ref with given path exists and returns error other than gorm.ErrRecordNotFound
+func IsRefExist(ctx context.Context, allocationID, path string) (bool, error) {
+	db := datastore.GetStore().GetTransaction(ctx)
+	ref := new(Ref)
+	err := db.Select("path").Where("allocation_id=? AND path=?", allocationID, path).First(ref).Error
```
peterlimg: Perhaps we can use db.Select("count (1) > 0") to avoid  unnecessary `ref` decoding.

code/go/0chain.net/core/common/utils.go
```diff
@@ -22,3 +25,15 @@ func ToKey(key interface{}) string {
...
+// getParentPaths For path /a/b/c.txt, will return [/a,/a/b]
+func GetParentPaths(fPath string) []string {
+	splittedPaths := strings.Split(fPath, "/")
+	var paths []string
+
+	for i := 0; i < len(splittedPaths); i++ {
+		subPath := strings.Join(splittedPaths[0:i], "/")
+		paths = append(paths, subPath)
+	}
+	return paths[2:]
```
peterlimg: This will panic if fPath is "". Please add unit tests to check the edge cases. 

Also, do you think we should consider the cases such as "/a//b", and //a//b, etc?

peterlimg (changes requested):

code/go/0chain.net/core/common/utils.go
lpoli: I've added `filepath.Clean()` in this function and removed it from `validateParentType` function.

For above cases; I think its necessary to cleaning path before processing. It should be done in functions as early as relevant and we should check against if filepath is not absolute.

peterlimg (approved):
lgtm

PR #586, opened 2022-03-15
fix(init): try ping postgres to check if it is ready

cnlangzi:
### Fixes
- add ping to check if postgres is ready. Because only tcp+port is checked on `gorm.Open`.  it is considered as ok if postgres is ready, but database is not initialized or login/password is incorrect. see detail bug on https://0chain.slack.com/archives/C02AV6MKT36/p1647005595089089?thread_ts=1646996040.723439&cid=C02AV6MKT36 . so the old code is not enough good to check it.

### Tests

### DIFF
--- a/code/go/0chain.net/blobbercore/datastore/postgres.go
+++ b/code/go/0chain.net/blobbercore/datastore/postgres.go
@@ -2,6 +2,8 @@ package datastore
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
 	"time"
 
@@ -26,6 +28,7 @@ func (store *postgresStore) Open() error {
 		config.Configuration.DBPassword)), &gorm.Config{
 		SkipDefaultTransaction: true, // https://gorm.io/docs/performance.html#Disable-Default-Transaction
 		PrepareStmt:            true, //https://gorm.io/docs/performance.html#Caches-Prepared-Statement
+		DisableAutomaticPing:   false,
 	})
 	if err != nil {
 		return common.NewErrorf("db_open_error", "Error opening the DB connection: %v", err)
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -26,6 +28,7 @@ func (store *postgresStore) Open() error {
 		config.Configuration.DBPassword)), &gorm.Config{
 		SkipDefaultTransaction: true, // https://gorm.io/docs/performance.html#Disable-Default-Transaction
 		PrepareStmt:            true, //https://gorm.io/docs/performance.html#Caches-Prepared-Statement
+		DisableAutomaticPing:   false,
```
peterlimg: we don't have to set it to `false` explicitly if it's false by default

peterlimg (changes requested):

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: fixed

code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -111,15 +117,15 @@ func (store *postgresStore) AutoMigrate() error {
 func (store *postgresStore) IsMigrated(m Migration) (bool, error) {
```
peterlimg: I don't see when will this function return true, can you check when should it return true?

peterlimg (changes requested):

code/go/0chain.net/blobbercore/datastore/postgres.go
cnlangzi: fixed

cnlangzi:
@peterlimg all comments are fixed. please check it again. thanks.

peterlimg (approved):
lgtm

PR #591, opened 2022-03-16
fix(challenge):db conn is not closed

### LINKED ISSUE #449: postgres connection might have memory leak issue
`OpenConnections always equals ConnectionsInUse`, it might be a memory leak. a db connection should be closed and backed to connection pool after command is executed.

Screen Shot 2021-12-02 at 08 05 55

cnlangzi:
### Fixes
- fixed https://github.com/0chain/blobber/issues/449

### Tests

-

### DIFF
--- a/code/go/0chain.net/blobbercore/challenge/challenge.go
+++ b/code/go/0chain.net/blobbercore/challenge/challenge.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/datastore"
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -41,58 +40,67 @@ func syncOpenChallenges(ctx context.Context) {
...
+
+	// it is not First and Next challenge
+	if !isValid {
+		logging.Logger.Error("[challenge]Challenge chain is not valid")
+		return
+	}
+
+	logging.Logger.Info("[challenge]add: ", zap.String("challenge_id", nextChallenge.ChallengeID))
+	nextChallenge.Status = Accepted
+	nextChallenge.CreatedAt = common.ToTime(nextChallenge.Created)
```
peterlimg, line 95: Shouldn't the UpdatedAt = common.ToTime(...), and the CreatedAt keep unchanged?

code/go/0chain.net/blobbercore/challenge/entity.go
```diff
@@ -107,6 +111,11 @@ func unMarshalField(stringObj datatypes.JSON, dest interface{}) error {
 }
 
 func (cr *ChallengeEntity) Save(ctx context.Context) error {
+	db := datastore.GetStore().GetTransaction(ctx)
+	return cr.SaveWith(db)
+}
+
+func (cr *ChallengeEntity) SaveWith(db *gorm.DB) error {
```
peterlimg, line 118: Seems the db is the only option for SaveWith, if so we should still use it the previous way to keep it simple

peterlimg (changes requested):

code/go/0chain.net/blobbercore/challenge/challenge.go
cnlangzi: No. it is new challenge from blockchain. so CreatedAt means when it is generated on blockchain, not is added on db.

code/go/0chain.net/blobbercore/challenge/entity.go
cnlangzi: NO. they are different.  it is a db session with heavy transaction if it is got from `GetTransaction(ctx)`. It is a normal db session without transaction in `SaveWith`.  that is why It is added. but `Save` is still used on other code with built-in transaction. it is not easy to refactor it. 
On blobber, transaction always be started in `WithConnection` and `WithReadonlyConnection` for a new incoming request. it is not good .but it is really difficult to refactor all of them.

cnlangzi:
@peterlimg all comments are updated. please check it again. thanks.

peterlimg (approved):
lgtm

PR #615, opened 2022-03-30
Enhancement/db automigration2

lpoli:
### Changes

    1. Add gorm's tags to help gorm automigrate schema from struct
    2. Add automigration function that drops existing tables and create new tables
    3. Provide options to keep or drop tables with dropAll option in .yaml file


code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -33,33 +34,56 @@ type AllocationChangeProcessor interface {
...
-	ConnectionID      string                      `gorm:"column:connection_id;primary_key"`
-	AllocationID      string                      `gorm:"column:allocation_id"`
-	ClientID          string                      `gorm:"column:client_id"`
-	Size              int64                       `gorm:"column:size"`
-	Changes           []*AllocationChange         `gorm:"ForeignKey:connection_id;AssociationForeignKey:connection_id"`
+	ConnectionID      string                      `gorm:"column:connection_id;primaryKey"`
+	AllocationID      string                      `gorm:"column:allocation_id;size:64;not null"`
+	ClientID          string                      `gorm:"column:client_id;size:64;not null"`
+	Size              int64                       `gorm:"column:size;not null;default:0"`
+	Changes           []*AllocationChange         `gorm:"foreignKey:CnxnID"`
```
peterlimg: CnxnID is a little bit confuse, does it mean connection id, if so, perhaps we can use connection_id or conn_id

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -33,33 +34,56 @@ type AllocationChangeProcessor interface {
...
 type AllocationChange struct {
-	ChangeID     int64  `gorm:"column:id;primary_key"`
-	Size         int64  `gorm:"column:size"`
-	Operation    string `gorm:"column:operation"`
-	ConnectionID string `gorm:"column:connection_id"`
-	Input        string `gorm:"column:input"`
+	ChangeID   int64                     `gorm:"column:id;primaryKey"`
+	Size       int64                     `gorm:"column:size;not null;default:0"`
+	Operation  string                    `gorm:"column:operation;size:20;not null"`
+	CnxnID     string                    `gorm:"column:connection_id;size:64;not null"`
```
peterlimg: ConnectionID would be easier to read.

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -31,27 +31,28 @@ const (
...
-	TotalSize      int64            `gorm:"column:size"`
-	UsedSize       int64            `gorm:"column:used_size"`
-	OwnerID        string           `gorm:"column:owner_id"`
-	OwnerPublicKey string           `gorm:"column:owner_public_key"`
-	RepairerID     string           `gorm:"column:repairer_id"` // experimental / blobber node id
-	PayerID        string           `gorm:"column:payer_id"`    // optional / client paying for all r/w ops
-	Expiration     common.Timestamp `gorm:"column:expiration_date"`
+	ID             string           `gorm:"column:id;size:64;primaryKey"`
+	Tx             string           `gorm:"column:tx;size:64;not null;unique;index:idx_unique_allocations_tx,unique"`
+	TotalSize      int64            `gorm:"column:size;not null;default:0"`
```
peterlimg, line 36: I believe the default value would be 0 by itself, do we need have the default:0 tag

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -31,27 +31,28 @@ const (
...
+	OwnerPublicKey string           `gorm:"column:owner_public_key;size:512;not null"`
+	RepairerID     string           `gorm:"column:repairer_id;size:64;not null"`
+	PayerID        string           `gorm:"column:payer_id;size:64;not null"`
+	Expiration     common.Timestamp `gorm:"column:expiration_date;not null"`
 	// AllocationRoot allcation_root of last write_marker
-	AllocationRoot   string        `gorm:"column:allocation_root"`
-	BlobberSize      int64         `gorm:"column:blobber_size"`
-	BlobberSizeUsed  int64         `gorm:"column:blobber_size_used"`
-	LatestRedeemedWM string        `gorm:"column:latest_redeemed_write_marker"`
+	AllocationRoot   string        `gorm:"column:allocation_root;size:64;not null;default:''"`
```
peterlimg, line 44: Default string will have value '', do we need to  set it explicitly with the tag?

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -31,27 +31,28 @@ const (
...
+	BlobberSize      int64         `gorm:"column:blobber_size;not null;default:0"`
+	BlobberSizeUsed  int64         `gorm:"column:blobber_size_used;not null;default:0"`
+	LatestRedeemedWM string        `gorm:"column:latest_redeemed_write_marker;size:64"`
 	IsRedeemRequired bool          `gorm:"column:is_redeem_required"`
-	TimeUnit         time.Duration `gorm:"column:time_unit"`
-	IsImmutable      bool          `gorm:"is_immutable"`
+	TimeUnit         time.Duration `gorm:"column:time_unit;not null;default:172800000000000"`
```
peterlimg, line 49: What's the default value 172800000000000 for

code/go/0chain.net/blobbercore/allocation/entity.go
```diff
@@ -31,27 +31,28 @@ const (
...
+	IsImmutable      bool          `gorm:"is_immutable;not null"`
 	// Ending and cleaning
-	CleanedUp bool `gorm:"column:cleaned_up"`
-	Finalized bool `gorm:"column:finalized"`
+	CleanedUp bool `gorm:"column:cleaned_up;not null;default:false"`
```
peterlimg, line 52: Isn't the bool value have the default value of false

code/go/0chain.net/blobbercore/automigration/automigration.go
```diff
@@ -0,0 +1,159 @@
...
+		for _, tbMdl := range tableModels {
+			if _, ok := tablesToKeep[tbMdl.TableName()]; ok {
+				continue
+			}
+			if err := db.Migrator().DropTable(tbMdl); err != nil {
+				return err
+			}
+		}
+	} else {
+		for _, tableName := range config.Configuration.DBTablesToDrop {
```
peterlimg: If tables exist in DBTablesToDrop and DBTablesToKeep, should they be dropped or kept?

code/go/0chain.net/blobbercore/datastore/postgres.go
```diff
@@ -18,6 +17,17 @@ type postgresStore struct {
...
 }
 
+func (p *postgresStore) GetPgDB() (*gorm.DB, error) {
+	return gorm.Open(postgres.Open(
+		fmt.Sprintf("host=%v port=%v user=%v password=%v sslmode=disable",
+			config.Configuration.DBHost,
+			config.Configuration.DBPort,
+			config.Configuration.PGUserName,
+			config.Configuration.PGPassword),
+	))
```
peterlimg, line 27: Any reason why the gorm.Config {SkipDefaultTransaction: true, and PrepareStmt: true} were not set, I saw they were used before.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -58,42 +58,58 @@ func (a *Attributes) Validate() (err error) {
...
 	DeletedAt gorm.DeletedAt `gorm:"column:deleted_at"` // soft deletion
 
-	ChunkSize int64 `gorm:"column:chunk_size" dirlist:"chunk_size" filelist:"chunk_size"`
+	ChunkSize int64 `gorm:"column:chunk_size;not null;default:65536" dirlist:"chunk_size" filelist:"chunk_size"`
+}
+
+// BeforeCreate Hook that gets executed to update create and update date
+func (ref *Ref) BeforeCreate(tx *gorm.DB) (err error) {
+	ref.CreatedAt = time.Now()
+	ref.UpdatedAt = time.Now()
```
peterlimg: ref.UpdatedAt = ref.CreatedAt

sql/00-create-user.sql
```diff
@@ -1,7 +0,0 @@
-CREATE extension ltree;
-CREATE DATABASE blobber_meta;
-
-\connect blobber_meta;
-CREATE USER blobber_user WITH ENCRYPTED PASSWORD 'blobber';
```
peterlimg, line 5: Does the deploy team know the password 'blobber' used here, if not please notify them to remember change the password when deploy to production env.

sql/02-create-indexes.sql
```diff
@@ -1,35 +0,0 @@
-\connect blobber_meta;
-BEGIN;
-DROP INDEX IF EXISTS idx_unique_allocations_tx;
-CREATE UNIQUE INDEX idx_unique_allocations_tx ON allocations (tx);
-
-DROP INDEX IF EXISTS idx_pendings_cab;
```
peterlimg, line 6: What does the `cab` mean, can we use meaningful name to make it easier to read, thanks.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/allocation/allocationchange.go
lpoli: We cannot use `connection_id` because of automigration issue. I will fall back to `ConnID`

code/go/0chain.net/blobbercore/allocation/entity.go
lpoli: This is the default value for sql. It will be used by gorm's automigration to set default value of column

code/go/0chain.net/blobbercore/allocation/entity.go
lpoli: This is the default value for sql. It will be used by gorm's automigration to set default value of column

code/go/0chain.net/blobbercore/allocation/entity.go
lpoli: This is weird, yes. I exactly tried to replicate every thing from .sql files. Hence this code.
We could make further changes in next PR after this one passes as it will be easier to make database-changes then.

@peterlimg  Should I try to make reasonable changes?

code/go/0chain.net/blobbercore/allocation/entity.go
lpoli: This is the default value for sql. It will be used by gorm's automigration to set default value of column

code/go/0chain.net/blobbercore/automigration/automigration.go
lpoli: It is based on `DBDropAllTables` value. If it is `true` it will take values from `DBTablesToKeep` otherwise from `DBTablesToDrop`.

This two lists are not overlapped.

code/go/0chain.net/blobbercore/datastore/postgres.go
lpoli: @peterlimg gorm's config is still used. 
This `db` client is only used to create `user` and `database`. After this, `db.DB().Close()` is called and then `db` is set to the one that you saw before.

```
func AutoMigrate(pgDB *gorm.DB) error {
	if err := createUser(pgDB); err != nil {
		return err
	}

	if err := createDB(pgDB); err != nil {
		return err
	}

	if err := grantPrivileges(pgDB); err != nil {
		return err
	}

	d, err := pgDB.DB()
	if err != nil {
		return err
	}

	if err := d.Close(); err != nil {
		return err
	}

	if err := datastore.GetStore().Open(); err != nil {
		return err
	}

	db := datastore.GetStore().GetDB()

	return migrateSchema(db)
}

```

sql/02-create-indexes.sql
lpoli: Previously they use client_id, allocation_id and blobber_id, shortform `cab`. Now we don't have blobber_id so I will rename it

lpoli: @peterlimg 
These all sql files are actually deleted and no more used. In the gorm tag I will rename it

Previously they use client_id, allocation_id and blobber_id, shortform cab. Now we don't have blobber_id so I will rename it

sql/00-create-user.sql
lpoli: I've put these values in `0chain_blobber.yaml` file. I will let them know.

lpoli:
https://github.com/0chain/system_test/runs/5773017122?check_suite_focus=true

System test has passed for this PR.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -33,33 +34,56 @@ type AllocationChangeProcessor interface {
...
-	ChangeID     int64  `gorm:"column:id;primary_key"`
-	Size         int64  `gorm:"column:size"`
-	Operation    string `gorm:"column:operation"`
-	ConnectionID string `gorm:"column:connection_id"`
-	Input        string `gorm:"column:input"`
+	ChangeID   int64                     `gorm:"column:id;primaryKey"`
+	Size       int64                     `gorm:"column:size;not null;default:0"`
+	Operation  string                    `gorm:"column:operation;size:20;not null"`
+	ConnID     string                    `gorm:"column:connection_id;size:64;not null"`
+	Connection AllocationChangeCollector `gorm:"foreignKey:ConnID"` // References allocation_connections(connection_id)
```
peterlimg: Seems `AlloocationChangeCollector` does not have `ConnID` field, so perhaps `ConnectionID`

peterlimg (changes requested):

code/go/0chain.net/blobbercore/allocation/allocationchange.go
lpoli: @peterlimg We cannot have `ConnectionID` as a field name when using gorm's automigration. It just doesn't work. This is issue with gorm but there is no other workaround.

But your concern with `ConnID` field; Gorm can create reference to the `allocation_connections` table. It actually is the gorm's tag that matters. And for gorm tag we have column name as `connection_id`.

code/go/0chain.net/blobbercore/allocation/entity.go
peterlimg: okay, leave it then, change if necessary in another PR

code/go/0chain.net/blobbercore/automigration/automigration.go
peterlimg: We'd better have code to check if there are tables overlapped, or define the priority about how to process it if overlapped. For example, suppose tables overlap is allowed, and  `DBTableToKeep` has higher priority, then we should exclude them from `DBTablesToDrop`, and drop the remaining ones only.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
peterlimg: Isn't the `foreignKey:ConnID` reference to the primary key defined in `AllocationChangeCollector` struct, which has `ConnectionID string` field with `gorm:"column:connection_id;primaryKey"` tag? I don't see the `ConnID` defined here, so I thought it mistakenly set it as `ConnID`. Anyway, do think the `foreignKey:ConnID` could reference it?

peterlimg (changes requested):

code/go/0chain.net/blobbercore/allocation/allocationchange.go
lpoli: If we use `ConnectionID`, automigration will panic.

`ConnID` is not needed to define in `AllocationChangeCollector` struct, but if you see `gorm's` tag for `ConnID`, its `connection_id`. 
It will only then create proper reference between tables.

code/go/0chain.net/blobbercore/automigration/automigration.go
lpoli: So this means we don't need to use `DBDropAllTables` field.

peterlimg: Yes, seems like the `DBDropAllTables` is not necessary. As the `DBTablesToKeep` and `DBTablesToDrop` can present the drop all case. 

One issue about **not** using `DBDropAllTables` is if we have a lot of tables, it's annoying to put all tables to the `DBTablesToDrop`, then the `DBDropAllTables` would be helpful, but that will need us to have the way to get all the tables, and then drop all of them.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -33,33 +34,56 @@ type AllocationChangeProcessor interface {
...
-	ConnectionID      string                      `gorm:"column:connection_id;primary_key"`
-	AllocationID      string                      `gorm:"column:allocation_id"`
-	ClientID          string                      `gorm:"column:client_id"`
-	Size              int64                       `gorm:"column:size"`
-	Changes           []*AllocationChange         `gorm:"ForeignKey:connection_id;AssociationForeignKey:connection_id"`
+	ConnectionID      string                      `gorm:"column:connection_id;primaryKey"`
+	AllocationID      string                      `gorm:"column:allocation_id;size:64;not null"`
+	ClientID          string                      `gorm:"column:client_id;size:64;not null"`
+	Size              int64                       `gorm:"column:size;not null;default:0"`
+	Changes           []*AllocationChange         `gorm:"foreignKey:ConnID"`
```
peterlimg: Why the foreighKey here is not `ChangeID`, isn't `ConnID` in `AllocationChange` references the AllocationChangeCollector itself?

code/go/0chain.net/blobbercore/allocation/allocationchange.go
peterlimg: Yes, I got what happens here, perhaps if we rename it to `AllocationChangeCollectorRefer` or `ConnectionRefer` would make it easier to understand. Check the tables.go I changed here: https://gist.github.com/peterlimg/2dbbb6916b1a46e7219c7cdd2e6880c4

peterlimg (changes requested):

lpoli:
I think `ConnID` is better than `ConnectionRefer` .

An alternative could be using `id` as column name. https://gist.github.com/lpoli/a04d38365be6bebb96ad2d444ee6be5f

code/go/0chain.net/blobbercore/allocation/allocationchange.go
lpoli: `ChangeID` is primary key of `allocation_changes` table. How would you use `preload` if `Changes` field has `ChangeID` to reference `allocation_changes` table? It should use field that refers to `connection_id` in `allocation_changes` table.

peterlimg:
> I think `ConnID` is better than `ConnectionRefer` .
> 
> An alternative could be using `id` as column name. https://gist.github.com/lpoli/a04d38365be6bebb96ad2d444ee6be5f

Yes, this one looks better

peterlimg:
@lpoli resolve the conflicts please, I believe this PR is ready to be merged once it passed the system tests.

peterlimg (approved):
lgtm

PR #642, opened 2022-04-20
Enhancement/disk utilization

### LINKED ISSUE #498: Track mount/unmount of disk on fly
As we are making progress on storage capacity of blobbers, it is also necessary to have track of mounted and unmounted disk in runtime. So for example, if new disk is mounted to the system for it to store allocation's data, this new capacity needs to be reflected in blockchain as well without blobber going offline. So this is similar to `bl-update` that we do in zboxcli.

We need some goroutine that keeps track of mounts/umounts for storage. It needs to update blobber's capacity to blockchain.

This is also required to have in Sharders as well. When some disk is mounted then it should be able to store blocks in this disk as well, without going offline.

lpoli:
### Changes
This PR manages files in blobber so that millions files won't get crowded in same directory.
Using fixed number of files to put in a directory is not feasible because allocation can update its size.

Like previous method, I have used random property of 256 hash of allocation and content hash. Allocation hash(id), to create parent directories for all files belonging to that allocation, and content hash to create directories and file name for some file.

Previous method had overuse of directories and number of directories that an allocation could have was hard coded.
It used content_hash's first 9 characters to create directory for a file.
So there could be 16^9 = 68719476736(68 billion directories).

Now blobber can change this number as per its capacity. 
Also in previous method, it would put all allocation in same directory i.e. /data/blobber/files. This would allow a directory to put millions allocation in a single directory thus making lookup inefficient. This also has been fixed.

### Fixes
1. Over crowd of millions files in same directory.
2. Use of lock to delete files as there can be two files having same content hash.
3. Data inconsistency due to file movement to minio. A ref gets updated for `on_cloud` field. If other refs too had been referring to same content hash then there is no way to get data for those refs.
4. Partially separate integration test and unit test. Later on this requires complete independence on each other.


.github/workflows/tests.yml
```diff
@@ -103,4 +103,4 @@ jobs:
           #sudo make integration-tests 
           go17=$(which go)
           root=$(pwd)
-          sudo CGO_ENABLED=1 root=$root integration=1 $go17 test -tags bn256  ./...
\ No newline at end of file
+          sudo CGO_ENABLED=1 root=$root integration=1 $go17 test -tags=integration  ./...
```
cnlangzi: please use `-tags="bn256 integtation"`

Makefile
```diff
@@ -56,7 +56,7 @@ local-run:
 	--hostname 127.0.0.1 \
 	--deployment_mode 0 \
 	--keys_file ../docker.local/keys_config/b0bnode1_keys.txt  \
-	--files_dir ./data/blobber/files \
+	--mount_point ./data/blobber/files \
```
cnlangzi: please keep it with `--files_dir`, if not. it will break all helm charts.

cnlangzi (dismissed):

Makefile
cnlangzi: or please update it on dockerfile , cli and helm charts too

code/go/0chain.net/blobbercore/filestore/file_manager.go
```diff
@@ -0,0 +1,495 @@
...
+		return
+	}
+
+	alloc.allocatedSize = uint64(allocatedSize)
+
+}
+
+func (fm *fileManager) getAllocation(allocID string) *allocation {
+	fm.rwMU.RLock()
+	defer fm.rwMU.Unlock()
```
peterlimg: do not use defer here, so as other places below

Makefile
lpoli: I think its better to update in cli and helm charts. For files storage, It must be mountpoint. So the flag name should be appropriate as well.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | enhancement/disk-utilization  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | enhancement/disk-utilization  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | enhancement/disk-utilization  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobber/config.go
```diff
@@ -40,16 +51,28 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.ChallengeResolveNumWorkers = viper.GetInt("challenge_response.num_workers")
 	config.Configuration.ChallengeMaxRetires = viper.GetInt("challenge_response.max_retries")
 
+	config.Configuration.AutomaticUpdate = viper.GetBool("disk_update.automatic_update")
+	blobberUpdateIntrv := viper.GetInt("disk_update.blobber_update_interval")
```
peterlimg: Change value type of `blobber_update_interval` to time.Duration, and use viper.GetDuration() to load it.

code/go/0chain.net/blobber/flags.go
```diff
@@ -47,10 +47,6 @@ func parseFlags() {
 	fmt.Print("> load flags")
 	flag.Parse()
```
peterlimg, line 49: We don't need to check if mountPoint is empty?

code/go/0chain.net/blobbercore/allocation/protocol.go
```diff
@@ -132,6 +134,13 @@ func VerifyAllocationTransaction(ctx context.Context, allocationTx string, reado
 	a.TimeUnit = sa.TimeUnit
 	a.IsImmutable = sa.IsImmutable
 
+	m := map[string]interface{}{
+		"allocation_id":  a.ID,
+		"allocated_size": (sa.Size + int64(len(sa.BlobberDetails)-1)) / int64(len(sa.BlobberDetails)),
+	}
+
+	go filestore.GetFileStore().UpdateAllocationMetaData(m)
```
peterlimg: What would happen if `UpdateAllocationMetaData()` failed, will we got inconsistent state of the allocated size?

code/go/0chain.net/blobbercore/datastore/mock_store.go
```diff
@@ -39,7 +39,9 @@ func MockTheStore(t *testing.T) sqlmock.Sqlmock {
 	var mock sqlmock.Sqlmock
 	var err error
 	db, mock, err = sqlmock.New()
-	require.NoError(t, err)
+	if t != nil {
```
peterlimg, line 42: This nil checking is unnecessary, we shall never call the function with nil `t`.

code/go/0chain.net/blobbercore/filestore/file_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	"sync"
+
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
+	"github.com/0chain/blobber/code/go/0chain.net/core/common"
+	"github.com/0chain/blobber/code/go/0chain.net/core/encryption"
+	"github.com/minio/minio-go"
+)
+
+type FileStore struct {
+	mp      string
```
peterlimg: Add appropriate comments so that user can understand what `mp` represents without checking the code below, or rename it to `mountPoint`.

code/go/0chain.net/blobbercore/filestore/file_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	mAllocs map[string]*allocation
+
+	allocMu *sync.Mutex
+	rwMU    *sync.RWMutex
+
+	mc                  *minio.Client
+	bucket              string
+	currentDiskCapacity uint64
```
peterlimg: If there's no other capacity, perhaps use diskCapacity is simpler

code/go/0chain.net/blobbercore/filestore/misc.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"errors"
+
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
+)
+
+// TODO make return values uint
+var getDirLevelsForAllocations = func() []int {
+	return []int{2, 1} // default
```
peterlimg, line 14: Can you add comments to describe how the DirLevels* works

code/go/0chain.net/blobbercore/filestore/state.go
```diff
@@ -0,0 +1,208 @@
...
+	defer cnCl()
+
+	ctx = datastore.GetStore().CreateTransaction(ctx)
+	db := datastore.GetStore().GetTransaction(ctx)
+	if db == nil {
+		return errors.New("could not get db client")
+	}
+
+	var allocations []*dbAllocation
+	if err := db.Model(&dbAllocation{}).Find(&allocations).Error; err != nil {
```
peterlimg: Will this load all the allocations?

peterlimg (changes requested):
Reviewed partially, I'll continue the code review later

code/go/0chain.net/blobber/flags.go
lpoli: It is checked in config.go because mount point can be given from config.yaml file as well.
Should we remove it from config.yaml file, then we need to check it here

code/go/0chain.net/blobbercore/filestore/state.go
lpoli: Yes it will.
Do you mean we should get allocations in batches?

code/go/0chain.net/blobbercore/datastore/mock_store.go
lpoli: I've used it calling in `init` functions. 
Should we put initialization in regular functions? Eg. You can see in renamefilechange_test.go

code/go/0chain.net/blobbercore/allocation/protocol.go
lpoli: It will ignore update if there is type mismatch. Otherwise it won't fail.
```
        m := map[string]interface{}{
		"allocation_id":  a.ID,
		"allocated_size": (sa.Size + int64(len(sa.BlobberDetails)-1)) / int64(len(sa.BlobberDetails)),
	}
```

With this map it will never fail.

One solution for this would be to avoid using go routine and return error, should it fail to update.
Do you suggest it?

code/go/0chain.net/blobber/config.go
lpoli: From user side, it would be easy to put integer value in minute or second. 
One second would be equivalent to `10^9` duration value.

I think it is better to keep it as it is.

code/go/0chain.net/blobbercore/filestore/file_manager.go
lpoli: Don't understand why. Can you please explain why it should not be used?

code/go/0chain.net/blobbercore/filestore/minio.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+
+	bucket = config.Configuration.MinioBucket
+	region := config.Configuration.MinioRegion
+
+	logging.Logger.Info(fmt.Sprintf("Checking if bucket %s exists", bucket))
+	isExist, err := mc.BucketExists(bucket)
+	switch {
+	case isExist:
```
peterlimg: Please check error first

code/go/0chain.net/blobbercore/filestore/minio.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return
+		}
+	}
+
+	return
+}
+
+func (fs *FileStore) MinioUpload(fileHash, filePath string) (err error) {
+	if fs.mc != nil {
```
peterlimg: If checking fs.nc is nil is needed, please check and return error early, so as the other methods below, e.g:

```go
if fs.mc == nil {
return common.NewError(...)
}

// continue the upload work
```

Also, is `common.NewError(` a must, if not we can return `errors.New("minio_client_not_set")`

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+		return false, common.NewError("invalid_hash", "calculated content hash does not match with client's content hash")
+	}
+
+	key := getKey(allocID, fileData.Hash)
+	l, _ := contentHashMapLock.GetLock(key)
+	l.Lock()
+
+	fileObjectPath, err := fs.GetPathForFile(allocID, fileData.Hash)
+	if err != nil {
+		return false, common.NewError("get_file_path_error", err.Error())
```
peterlimg, line 202: lock is not released when error happens

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+
+	err = createDirs(filepath.Dir(fileObjectPath))
+	if err != nil {
+		return false, common.NewError("blob_object_dir_creation_error", err.Error())
```
peterlimg, line 207: lock is not released

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+
+	alloc.tmpMU.Lock()
+	defer alloc.tmpMU.Unlock()
+
+	if size < 0 {
+		alloc.tmpFileSize -= uint64(size)
+	} else {
+		alloc.tmpFileSize += uint64(size)
+	}
```
peterlimg: Since size is signed value, we can do `alloc.tmpFileSize += size` instead of doing extra type casting

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+		logging.Logger.Warn("invalid_path", zap.String("fileObjectPath", fileObjectPath), zap.Error(err))
+		return err
+	}
+
+	fs.updateAllocTempFileSize(allocID, -size)
+
+	return nil
+}
+
+func (fs *FileStore) GetFileBlock(allocID string, fileData *FileInputData, blockNum, numBlocks int64) ([]byte, error) {
```
peterlimg, line 298: Add comments to indicate that the blockNum starts from 1, otherwise user may misuse it.

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+			return nil, nil, common.NewError("minio_download_failed", "Unable to download from minio with err "+err.Error())
+		}
+		file, err = os.Open(fileObjectPath)
+		if err != nil {
+			return nil, nil, err
+		}
+	case err != nil:
+		return nil, nil, err
+	}
+	defer file.Close()
```
peterlimg, line 377: The code above between 349 and 371 is basically the same as code in GetFileBlock(), please extract them as a separate method for getting or download file to reduce duplicate code.

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+
+	var returnBytes []byte
+
+	fi, err := file.Stat()
+	if err != nil {
+		return nil, nil, common.NewError("stat_error", err.Error())
+	}
+
+	totalChunks := int(math.Ceil(float64(fi.Size()) / float64(fileData.ChunkSize)))
```
peterlimg, line 386: Could return early if the totalChunks is 0

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+	var err error
+	defer func() {
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	tempDir := fstr.getAllocTempDir(ID)
+
+	finfo, err := os.Stat(tempDir)
```
peterlimg, line 491: The error will shadow the error above in defer func()

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+// getPath returns "/" separated strings with the given levels.
+func getPath(hash string, levels []int) string {
+	var count int
+	var pStr []string
+	for _, i := range levels {
+		pStr = append(pStr, hash[count:count+i])
+		count += i
+	}
+	pStr = append(pStr, hash[count:])
+	return strings.Join(pStr, "/")
```
peterlimg, line 554: Use filepath.Join

code/go/0chain.net/blobbercore/handler/calculate_hash_integration_test.go
```diff
@@ -12,6 +13,10 @@ import (
 )
 
 func TestBlobberGRPCService_CalculateHash(t *testing.T) {
+	if os.Getenv("integration") != "1" {
```
peterlimg: Put this integration test checking in a helper function, such as:
```go
func isIntegrationEnabled(t *testing) {
    // 
}
```

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+	finfo, err = f.Stat()
+	if err != nil {
+		return nil, common.NewError("file_stat_error", err.Error())
+	}
+
+	fileRef := &FileOutputData{}
+
+	currentSize := finfo.Size()
+	if currentSize > initialSize { // Is chunk new or rewritten
+		fileRef.ChunkUploaded = true
```
peterlimg, line 102: If the file size is less than chunk size, should we set ChunkUploaded to true?

code/go/0chain.net/blobbercore/handler/file_command_update.go
```diff
@@ -91,7 +91,7 @@ func (cmd *UpdateFileCommand) ProcessContent(ctx context.Context, req *http.Requ
 	allocationSize := connectionObj.Size
 
 	// only update connection size when the chunk is uploaded by first time.
-	if !fileOutputData.ChunkUploaded {
+	if fileOutputData.ChunkUploaded {
```
peterlimg, line 105: The comment of 'only update connection size when the chunk is uploaded by first time.' seems not true as each time when new content is uploaded, the ChunkUploaded is true, and the allocationSize will be updated then.

code/go/0chain.net/blobbercore/handler/tests_common.go
```diff
@@ -0,0 +1,189 @@
...
+	"strings"
+	"testing"
+
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/filestore"
+	"github.com/0chain/gosdk/core/util"
+	"github.com/0chain/gosdk/core/zcncrypto"
+	"github.com/0chain/gosdk/zcncore"
+)
+
+func setup(t *testing.T) {
```
peterlimg: Put tests related code in *_test.go

code/go/0chain.net/blobbercore/handler/worker.go
```diff
@@ -177,55 +182,68 @@ func startMoveColdDataToCloud(ctx context.Context) {
...
 	}
 }
 
 func moveFileToCloud(ctx context.Context, fileRef *reference.Ref) {
-	fs := filestore.GetFileStore()
-	allocation, err := fs.SetupAllocation(fileRef.AllocationID, true)
+	fileObjectPath, err := filestore.GetFileStore().GetPathForFile(fileRef.AllocationID, fileRef.ContentHash)
 	if err != nil {
-		Logger.Error("Unable to fetch allocation with error", zap.Any("allocationID", fileRef.AllocationID), zap.Error(err))
+		logging.Logger.Error("Error while getting path of file", zap.Error(err))
```
peterlimg, line 194: Return if error happens

code/go/0chain.net/blobbercore/handler/worker.go
```diff
@@ -177,55 +182,68 @@ func startMoveColdDataToCloud(ctx context.Context) {
...
+
+	var contentHashSharingRefsCount int64
+	condRef := reference.Ref{AllocationID: fileRef.AllocationID, ContentHash: fileRef.ContentHash}
+	if err := db.Model(&reference.Ref{}).Where(&condRef).Count(&contentHashSharingRefsCount).Error; err != nil {
+		logging.Logger.Error("Failed to get count", zap.Error(err))
+		return
+	}
 
-	if config.Configuration.ColdStorageDeleteLocalCopy {
+	if contentHashSharingRefsCount < 2 && config.Configuration.ColdStorageDeleteLocalCopy {
```
peterlimg: Can you comment about the magic number 2, otherwise we have to guess the meaning base on the context.

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+
+func (fs *FileStore) GetTempFilesSizeByAllocation(allocID string) uint64 {
+	alloc := fs.mAllocs[allocID]
+	if alloc != nil {
+		return alloc.tmpFileSize
+	}
+	return 0
+}
+
+func (fs *FileStore) GetTotalPermFilesSizeByAllocations() (s uint64) {
```
peterlimg: The Perm may confuse users, also if FileStore is not indexed by something else other than Allocations, we don't need the *ByAllocations postfix, so maybe use `GetTotalCommitFilesSize`

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,628 @@
...
+
+func (fs *FileStore) GetPermFilesSizeByAllocation(allocID string) uint64 {
+	alloc := fs.mAllocs[allocID]
+	if alloc != nil {
+		return alloc.filesSize
+	}
+	return 0
+}
+
+func (fs *FileStore) GetTotalFilesSizeByAllocations() (s uint64) {
```
peterlimg: Drop the postfix `*ByAllocations`

peterlimg (changes requested):

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: tmpFileSize is of type uint64. I kept it uint64 because it cannot be negative field.

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: @peterlimg They are not same. Even the purpose of function parameters are different.
I think `GetMerkleTree` would be proper name, right?

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: But totalChunks is never 0. If file exists, then it has size. it cannot be zero size

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: How do you mean it? It is the same err variable that is used in defer function.

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: Renamed function to `getPartialPath`.

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: First thing is we are now adding zero padding and not uploading file lesser than chunksize.

Even if file size is less than chunk size, then it should set `ChunkUploaded` to `true`.

code/go/0chain.net/blobbercore/handler/file_command_update.go
lpoli: Yes the comment is mistake. I will remove it.

code/go/0chain.net/blobbercore/handler/worker.go
lpoli: Have put condition `<= 1` instead of `< 2`. I had done `< 2`, as it is single instruction than for `<=1 `.
Anyway readability should not be forgotten.

code/go/0chain.net/blobbercore/filestore/file_manager.go
peterlimg: Defer has some performance consideration, and for this three line of code, it could be done in the following way without defer and having reasonable readability. Anyway, it's not a big problem, but I would recommend to do it without defer.

```go
func (fm *fileManager) getAllocation(allocID string) (alloc *allocation) {
    fm.rwMu.RLock()
    alloc = fs.mAllocs[allocID]
    fm.rwMu.RUnlock()
    return
}
```

code/go/0chain.net/blobber/config.go
peterlimg: User can set `1m` or `1s` for 1 minute or 1 second in yaml file, the viper.GetDuration() will parse the string to time.Duration for us.

code/go/0chain.net/blobbercore/allocation/protocol.go
peterlimg: Yes, I suggest to do it without goroutine if it is not a resources heavy process.

code/go/0chain.net/blobbercore/datastore/mock_store.go
peterlimg: Never mind, do it in regular function may complex the process, let's keep this nil checking

code/go/0chain.net/blobbercore/filestore/state.go
peterlimg: Just worried about the performance when we have large number of allocations, so as the RAM resources it would use. Perhaps do it in batches

code/go/0chain.net/blobbercore/filestore/storage.go
peterlimg: Do you mean the `fi.Size()` would never be less than the `fileData.ChunkSize` in line 380, if so we don't need to check zero value.

code/go/0chain.net/blobbercore/filestore/storage.go
peterlimg: Check the line 483, `finfo, err := os.State(tempDir)`, if we got err here and return in line 488, we expect that the error will be caught in the defer function above, but since the err variable is an new assigned value, hence the err in defer would still be nil, no error will be reported or panic.

peterlimg: Never mind, my bad.

code/go/0chain.net/blobbercore/filestore/storage.go
peterlimg: Oh as we will have padding zero so the file size would not be less than the fileData.ChunkSize, so should be fine

code/go/0chain.net/blobbercore/filestore/storage.go
peterlimg: It's for getting data blocks for challenge, not for MerkleTree calculation purpose, can you change it back to the previous name as its more obvious. Otherwise we may misuse it.

code/go/0chain.net/blobbercore/filestore/storage.go
```diff
@@ -0,0 +1,638 @@
...
+	}
+
+	err = createDirs(filepath.Dir(fileObjectPath))
+	if err != nil {
+		return false, common.NewError("blob_object_dir_creation_error", err.Error())
+	}
+
+	//move file from tmp location to the objects folder
+	err = os.Rename(tempFilePath, fileObjectPath)
+	l.Unlock()
```
peterlimg: The l.Unlock() will be called twice if the os.Rename() return err, as the defer function above will call it again.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: currently it can be less than chunksize. However `math.Ceil` would raise the value to 1.

code/go/0chain.net/blobbercore/filestore/storage.go
lpoli: @peterlimg No it won't create new error variable. It will be the same err. It will only create new `finfo` variable.

Check this out:
```
package main

import (
	"errors"
	"fmt"
)

func main() {
	var err error
	defer func() {
		if err != nil {
			panic(err)
		}
	}()

	i, err := 1, errors.New("hey")
	fmt.Println(i)
}

```

code/go/0chain.net/blobbercore/filestore/state.go
```diff
@@ -0,0 +1,210 @@
...
+
+	err := db.Model(&dbAllocation{}).FindInBatches(&allocations, 1000, func(tx *gorm.DB, batch int) error {
+		for _, alloc := range allocations {
+			a := allocation{
+				allocatedSize: uint64(alloc.BlobberSize),
+				mu:            &sync.Mutex{},
+				tmpMU:         &sync.Mutex{},
+			}
+
+			fs.setAllocation(alloc.ID, &a)
```
peterlimg: We can add a new method say `fs.setAllocations()` to add all allocations in one calling within one lock/unlock block.

peterlimg (changes requested):

peterlimg (approved):
lgtm

PR #676, opened 2022-05-10
Limit the combined directories and files per allocation

### LINKED ISSUE #639: Limit the number of dirs and files an allocation could have
This is (perhaps) a temporary fix for resolving the slow allocation root calculation when there are enormous dirs and files in an allocation. Issue #627 detailed the problem, but we may not consider to do it in splitting tables way. Instead, we will consider to optimize it by organization the dirs and files in MPT tree, discussed [here](https://0chain.slack.com/archives/GREHVKQKS/p1649465569155379), just like what we do for blockchain state. Anyway, we will not put effects on it before mainnet. What we currently need to do is limit the number of dirs and files each allocation could have. The steps could be:

- [ ] Add `max_dirs_files` to config file
- [ ] Check the count of the files and dirs of an allocation each time when adding new files, return error if the count reached the max allowed number

avanaur:
### Changes
Add new `max_dirs_files` config (default 1000) to limit the number of files and directories per allocation.

The check is applied on file actions that will create new file such as
- create directory
- file upload
- copy file
- new file (think same as file upload)

The check occurs first before any file ref changes. It is not too restrictive in a sense that it will not consider how many new refs a file action will make. As long as current total combined files and dirs are below limit, the whole file action will be allowed to go.

### Fixes
Fixes https://github.com/0chain/blobber/issues/639

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


peterlimg (approved):
lgtm

code/go/0chain.net/blobber/config.go
```diff
@@ -104,8 +104,8 @@ func setupConfig(configDir string, deploymentMode int) {
 
 	config.Configuration.MaxAllocationDirFiles =
 		viper.GetInt("max_dirs_files")
-	if config.Configuration.MaxAllocationDirFiles <= 0 {
```
cnlangzi: I would suggest to leave 0 as default. `0` is unlimited. it is better backwards compatible. It must be configured explicitly if blobber wants a hard limitation on it.

cnlangzi (changes requested):

code/go/0chain.net/blobber/config.go
peterlimg: @cnlangzi @avanaur sorry but I have to say the default is 50k, we want to apply the limit by default, otherwise we will have to notify blobbers to change their limit size. Unlimited allocation size is actually not an option.

cnlangzi (approved):

service-0chain:
Manual system tests [cancelled] with the following config
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #678, opened 2022-05-11
Fix/limit thumbnail size

lpoli:
### Changes
Delete object function has been optimized by removal redundant of logic and addition of goroutine for disk operations.

### Fixes
Thumbnail file size is limited to 1MB

### DIFF
--- a/code/go/0chain.net/blobbercore/allocation/deletefilechange.go
+++ b/code/go/0chain.net/blobbercore/allocation/deletefilechange.go
@@ -71,12 +71,17 @@ func (nf *DeleteFileChange) CommitToFileStore(ctx context.Context) error {
 		FindInBatches(&results, 100, func(tx *gorm.DB, batch int) error {
 
 			for _, res := range results {
-				limitCh <- struct{}{}
-				wg.Add(1)
 				var count int64
 				tx.Model(&reference.Ref{}).
 					Where("allocation_id=? AND content_hash=?", nf.AllocationID, res.ContentHash).
 					Count(&count)
+
+				if count != 0 {
+					continue
+				}
+
+				limitCh <- struct{}{}
+				wg.Add(1)
 
 				go func(res Result, count int64) {
 					defer func() {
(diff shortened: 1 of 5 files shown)


code/go/0chain.net/blobbercore/allocation/deletefilechange.go
```diff
@@ -58,20 +56,64 @@ func (nf *DeleteFileChange) DeleteTempFile() error {
...
+					Where("allocation_id=? AND content_hash=?", nf.AllocationID, res.ContentHash).
+					Count(&count)
+
+				go func(res Result, count int64) {
+					defer func() {
+						<-limitCh
+						wg.Done()
+					}()
+
+					if count == 0 {
```
peterlimg: We can do this count checking out of the gorountine, so if it's not 0, we don't need to start the goroutine.

code/go/0chain.net/blobbercore/handler/file_command_update.go
```diff
@@ -54,6 +58,14 @@ func (cmd *UpdateFileCommand) IsValidated(ctx context.Context, req *http.Request
 		return common.NewError("invalid_operation", "Operation needs to be performed by the owner, collaborator or the payer of the allocation")
 	}
 
+	_, thumbHeader, _ := req.FormFile(UploadThumbnailFile)
+	if thumbHeader != nil {
+		if thumbHeader.Size > MaxThumbnailSize {
+			return common.NewError("max_thumbnail_size",
+				fmt.Sprintf("Thumbnail size %d should not be greater than %d", thumbHeader.Size, MaxThumbnailSize))
```
peterlimg: We'd better not capitalize error messages, as they are usually printed following other context.

code/go/0chain.net/blobbercore/reference/object.go
```diff
@@ -3,77 +3,48 @@ package reference
...
-		found := false
-		for _, child := range dirRef.Children {
-			if child.Name == tSubDirs[treelevel] && child.Type == DIRECTORY {
-				dirRef = child
-				dirRef.HashToBeComputed = true
+
+	for _, subDir := range subDirs {
+		var found bool
+		for _, ref := range dirRef.Children {
+			if ref.Name == subDir && ref.Type == DIRECTORY {
```
peterlimg, line 33: Perhaps we should break immediately if the dirRef.Children do not have the subDir, basically this checking all start from the root path, if any level of the path do not exist in the corresponding dirRef, there must be something wrong.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/reference/object.go
lpoli: It does break with `return` statement.

peterlimg (approved):
lgtm

cnlangzi (commented):
@peterlimg  @lpoli , this PR breaks multiple things
- breaks 204 status for partial delete
- it is incorrect to delete thumbnail if hash is not empty. Because hash always has value even it is empty.

PR #679, opened 2022-05-13
Delete thumbnail even if file is not to be deleted

lpoli:
### Fixes
Even if there are content sharing `refs`, they are not supposed to share `thumbnail`. So if delete operation is called, thumbnail should be deleted regardless

### DIFF
--- a/code/go/0chain.net/blobbercore/allocation/deletefilechange.go
+++ b/code/go/0chain.net/blobbercore/allocation/deletefilechange.go
@@ -9,10 +9,15 @@ import (
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/datastore"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/filestore"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/reference"
+	"github.com/0chain/blobber/code/go/0chain.net/core/common"
 	"github.com/0chain/blobber/code/go/0chain.net/core/logging"
 	"gorm.io/gorm"
 
 	"go.uber.org/zap"
+)
+
+var (
+	NoNewLock = common.NewError("not_new_lock", "")
 )
 
 type DeleteFileChange struct {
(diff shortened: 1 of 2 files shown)


code/go/0chain.net/blobbercore/allocation/deletefilechange.go
```diff
@@ -76,7 +81,7 @@ func (nf *DeleteFileChange) CommitToFileStore(ctx context.Context) error {
 					Where("allocation_id=? AND content_hash=?", nf.AllocationID, res.ContentHash).
 					Count(&count)
 
-				if count != 0 {
+				if count != 0 && res.ThumbnailHash == "" {
```
cnlangzi, line 84: I would suggest to use `res.ThumbnailSize ==0` instead of `res.ThumbnailHash == ""`

code/go/0chain.net/blobbercore/allocation/deletefilechange.go
```diff
@@ -89,10 +94,12 @@ func (nf *DeleteFileChange) CommitToFileStore(ctx context.Context) error {
...
-							zap.String("content_hash", res.ContentHash))
+					if count == 0 {
+						err := filestore.GetFileStore().DeleteFile(nf.AllocationID, res.ContentHash)
+						if err != nil {
+							logging.Logger.Error(fmt.Sprintf("Error while deleting file: %s", err.Error()),
+								zap.String("content_hash", res.ContentHash))
+						}
 					}
 
 					if res.ThumbnailHash != "" {
```
cnlangzi, line 105: I would suggest to use `res.ThumbnailSize > 0` instead of `res.ThumbnailHash != ""`

cnlangzi (changes requested):

code/go/0chain.net/blobbercore/allocation/deletefilechange.go
lpoli: @cnlangzi The bug you are worrying does not exist now. So I think its enough to check if thumbnail hash is empty

cnlangzi (approved):
LGTM

peterlimg (approved):
lgtm

PR #705, opened 2022-05-31
Skip automigrating for existing tables

lpoli:
### Fixes
This PR add version table so that we can later on migrate tables and its data based on this information.
Also, the automigration will drop tables only if `automigration: true`, default value is `false`.


cnlangzi:
I don't how it works if a table is configured as keep_tables on yaml file, and there are new columns should be added, or some old columns should be dropped. let's waiting for @peterlimg  to review

code/go/0chain.net/blobbercore/automigration/automigration.go
```diff
@@ -129,20 +134,9 @@ func grantPrivileges(db *gorm.DB) error {
...
-	}
-
 	var migratingTables []tableNameI
 	for _, tblMdl := range tableModels {
 		tableName := tblMdl.TableName()
-		if _, ok := tablesToKeep[tableName]; ok {
-			continue
-		}
-
 		err := db.Migrator().DropTable(tableName)
```
cnlangzi: We need `Migration`. `drop` and `create` is `recreate` with fresh data, is not `schema migration`.

config/0chain_blobber.yaml
```diff
@@ -97,18 +97,12 @@ pg:
   user: postgres
   password: postgres
 db:
-  automigrate: true # true to automigrate with gorm's Automigration
+  automigrate: false # true to automigrate with gorm's Automigration
```
cnlangzi: it is unnecessary. AutoMigrate should always be checked. gorm will compare schema between Models and database, and build upgrade scripts for new tables, new columns, changed columns.....

AutoMigrate is `migrate schema without losing data`.

cnlangzi (changes requested):
I think you have some misunderstand about `AutoMigrate` on schema. why a table has to be dropped before migrating its schema?  nobody will drop table to loss all data when it is migrated. I can't understand your idea.

config/0chain_blobber.yaml
lpoli: @cnlangzi you got stale view of the PR.
I had already made the changes that you have requested.

code/go/0chain.net/blobber/config.go
```diff
@@ -132,6 +131,10 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.WebsiteUrl = viper.GetString("website_url")
 	config.Configuration.LogoUrl = viper.GetString("logo_url")
 	config.Configuration.Description = viper.GetString("description")
+	config.Configuration.Version = viper.GetString("version")
```
cnlangzi: add `version` in yaml file, it will have the same issue with keep_tables. it still has a deployment issue for blobber owner.

cnlangzi (changes requested):

config/0chain_blobber.yaml
cnlangzi: sorry, github view doesn't refresh properly.

config/0chain_blobber.yaml
```diff
@@ -1,4 +1,4 @@
-version: 1.0
+version: "1.0"
```
cnlangzi, line 1: it is configured for nothing. it is better to remove it for now.

cnlangzi (changes requested):

cnlangzi (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/automigrate-existing-table-issue  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #735, opened 2022-06-21
Fix blobber compilation for conductor tests

fabioDMFerreira:
Make writemarker and hashnode functions available to the integration tests bundle.

While running conductor tests we are getting next errors because the functions referred to are in a file that is not compiled on conductor tests.
```
Error: ../blobbercore/handler/handler.go:81:64: undefined: LockWriteMarker
Error: ../blobbercore/handler/handler.go:82:77: undefined: UnlockWriteMarker
Error: ../blobbercore/handler/handler.go:84:61: undefined: LoadRootHashnode
```


cnlangzi (commented):
@fabioDMFerreira please fix ci errors first

cnlangzi (approved):

PR #741, opened 2022-07-01
fix(challenge): use channel to trigger challenge worker instantly

cnlangzi:
### Changes
-

### Fixes
-  use channel to trigger next challenge worker instantly. 

### Tests

- 0chain:


code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -52,12 +54,11 @@ func syncOpenChallenges(ctx context.Context) {
 	}
 
 	for _, challengeObj := range blobberChallenges.Challenges {
-
 		if challengeObj == nil || challengeObj.ChallengeID == "" {
 			logging.Logger.Info("[challenge]open: No challenge entity from the challenge map")
 			continue
 		}
-
+		log.Println("challenge: sync", challengeObj.ChallengeID)
```
peterlimg: Let's use logging

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -169,6 +185,7 @@ func commitProcessed(ctx context.Context) {
 		for _, challenge := range challenges {
 			c := challenge
 			swg.Add()
+			log.Println("challenge: commit ", time.Now())
```
peterlimg: Use logging

peterlimg (changes requested):

code/go/0chain.net/blobbercore/challenge/worker.go
```diff
@@ -30,6 +37,8 @@ func startProcessAccepted(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			return
+		case next := <-waitToProcess:
```
ssardana08, line 40: do we still needs to be called from `processAccepted` ?

code/go/0chain.net/blobbercore/challenge/worker.go
```diff
@@ -14,11 +14,18 @@ func SetupWorkers(ctx context.Context) {
...
+	waitToCommit  = make(chan *ChallengeEntity)
+)
+
 func startCommitProcessed(ctx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
+		case next := <-waitToCommit:
+			commitChallenge(next)
```
ssardana08, line 28: do we still needs to be called from `commitProcessed` ?

ssardana08 (changes requested):

code/go/0chain.net/blobbercore/challenge/worker.go
cnlangzi: @sum-it08  Yes. it is keep challenge will be lost to process even blobber crashes and restarts.

code/go/0chain.net/blobbercore/challenge/worker.go
cnlangzi: @sum-it08  Yes. it is keep challenge will be lost to commit even blobber crashes and restarts.

code/go/0chain.net/blobbercore/challenge/challenge.go
cnlangzi: fixed

code/go/0chain.net/blobbercore/challenge/challenge.go
cnlangzi: fixed

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -131,7 +135,11 @@ func processAccepted(ctx context.Context) {
 }
 
 func validateChallenge(swg *sizedwaitgroup.SizedWaitGroup, challengeObj *ChallengeEntity) {
```
ssardana08, line 137: for this to be called we should just be playing with `challenge_response.frequency`
make it shorter if we need it to be called frequently

ssardana08 (commented):
@cnlangzi @peterlimg I don't think this PR would be adding much value, but making challenge process more complex
We have a single config `challenge_response.frequency` which triggers process, validate, commit delay of challenge. We can just make that value small to process challenges faster. That's all. We don't need these new changes to make the challenge process faster

ssardana08:
changes to `challenge_response.frequency` should be able to make the challenge process fast.
We can close this for now

PR #751, opened 2022-07-12
File Rate Limiter

### LINKED ISSUE #671: request limit on CRUD operations
add a limit on the number of CRUD operations that the user can perform on an allocation to prevent Byzantine attacks.

avanaur:
### Changes

File Rate Limiter on upload and download.
When `file_rate_limiter` = 0, rate limiter is disabled.

As part of changes, `TryParseForm()` and `GetField()` were moved to common pkg

### Fixes
Resolves https://github.com/0chain/blobber/issues/671

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/core/common/rate_limiter.go
```diff
@@ -62,11 +79,87 @@ func UserRateLimit(handler ReqRespHandlerf) ReqRespHandlerf {
...
+		return baseKeys[0]
+	}
+
+	// OR just empty
+	baseKeys[0] = append(baseKeys[0], "")
+	return baseKeys[0]
+}
+
+func requestField(r *http.Request, key string) (string, bool) {
+	if r.Form != nil {
```
peterlimg: We can get URL query data from r.Form as well, so don't have to call the r.URL.Query below, and the code could be like:

```go
if err := r.ParseForm(); err != nil {
  return "", err
}

return r.Form.Get(key), nil
```

config/0chain_blobber.yaml
```diff
@@ -69,6 +69,7 @@ block_worker: http://198.18.0.98:9091
 
 handlers:
   rate_limit: 0 # 10 per second . it can't too small one if a large file is download with blocks
+  file_rate_limit: 1000 # 100 files per second
```
peterlimg: 1000 is too fast for both upload/download by default, the blobber would be under DDoS, so perhaps 100. And if the upload/download process get affected, then we might need to optimize the block size to reduce the requests number.

peterlimg (changes requested):

config/0chain_blobber.yaml
avanaur: Make sense. Adjust the chunk size instead of creating too many API requests for upload, and block size for download.

code/go/0chain.net/core/common/rate_limiter.go
avanaur: I opted to just moved `TryParseForm()` and `GetField()` from handler package to common and reuse that `GetField()` instead.

peterlimg (approved):
lgtm

PR #774, opened 2022-07-31
Fix/challenge issue

lpoli:
### Fixes
1. Added lock between object path calculation and writemarker commit in handler.CommitWrite function
2. Fix database connection leakage 
3. Use goroutines to send request to validators
4. Allow blank allocation to be valid.

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -81,6 +84,10 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
 		return err
 	}
 
+	// Lock allocation changes from happening in handler.CommitWrite function
+	allocMu := lock.GetMutex(allocationObj.TableName(), allocationObj.ID)
+	allocMu.Lock()
+
```
ssardana08, line 97: defer allocMu.Unlock()

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +173,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
 		postData["chunk_size"] = objectPath.ChunkSize
 	}
 
+	allocMu.Unlock()
```
ssardana08, line 189: it should be deferred just after locking

ssardana08 (changes requested):

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: @sum-it08 deferring is not good choice here. Unlock should be called as early as possible because it would be locking `handler.CommitWrite` which makes allocation changes.

Deferring will not unlock until it reaches `return` statement.

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -70,12 +71,14 @@ func saveNewChallenge(c *ChallengeEntity, ctx context.Context) {
 		}
 	}()
 
-	db := datastore.GetStore().GetDB()
-	if Exists(db, c.ChallengeID) {
+	ctx = datastore.GetStore().CreateTransaction(ctx)
```
cnlangzi: db.Transaction is enough good to do it. It is unnecessary to box/unbox db transaction in context.Context. There are no any memory/db leak.  commit/rollback will be executed automatically.  see https://gorm.io/docs/transactions.html#Transaction

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -70,12 +71,14 @@ func saveNewChallenge(c *ChallengeEntity, ctx context.Context) {
...
+	tx := datastore.GetStore().GetTransaction(ctx)
+
+	if Exists(tx, c.ChallengeID) {
 		return
```
cnlangzi: it makes transaction is never closed if challenge exists on db.old [db.Transaction](https://gorm.io/docs/transactions.html#Transaction) implementation is better.

cnlangzi (dismissed):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/challenge-issue  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/challenge-issue  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Yes this was the reason why tests were failing.
Fixed it with defer function

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Yes they both are the same thing. However, everywhere there is use of context to create and get new transaction.

cnlangzi: It is a **BAD** example. please don't follow it -_-!!!. 

By the way, almost of challenges are pulled more than times before it is committed. so it will waste a lot of db resource to create transaction for `Exists ` query only. 

It was fixed. but this PR will get it backed again now. 

I would suggest to revert current changes. it does nothing. there was not any db connection lake .

lpoli: It makes sense what you say.
We need to be careful to have db connections less than allocated by database.

If sometimes it uses `datastore.GetStore().GetDB()` for any query more than total db connections it will fail. But given that we use sized wait group it seems to be unlikely.
With creating transaction, we never will run out out connections issue.

I've reverted to what you have suggested.

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -117,56 +121,68 @@ func processAccepted(ctx context.Context) {
 		}
 	}()
 
-	db := datastore.GetStore().GetDB()
+	ctx = datastore.GetStore().CreateTransaction(ctx)
+	tx := datastore.GetStore().GetTransaction(ctx)
+
+	var err error
+	defer func() {
```
ssardana08: this can be made a function which can be deferred from all required function.

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -117,56 +121,68 @@ func processAccepted(ctx context.Context) {
...
+		tx.Commit()
+	}()
+
 	challenges := make([]*ChallengeEntity, 0)
-	db.Where(ChallengeEntity{Status: Accepted}).Find(&challenges)
-	if len(challenges) > 0 {
-		swg := sizedwaitgroup.New(config.Configuration.ChallengeResolveNumWorkers)
-		for _, c := range challenges {
-			logging.Logger.Info("[challenge]process: ",
+	tx.Where(ChallengeEntity{Status: Accepted}).Find(&challenges)
```
ssardana08: why did we change this from db to transaction fetch?

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -117,56 +121,68 @@ func processAccepted(ctx context.Context) {
...
-	if err := c.LoadValidationTickets(ctx); err != nil {
-		logging.Logger.Error("[challenge]validate: ",
-			zap.Any("challenge_id", c.ChallengeID),
-			zap.Time("created", c.CreatedAt),
-			zap.Error(err))
-		db.Rollback()
-		return
-	}
+	var err error
+	defer func() {
```
ssardana08: make this function and to be used everywhere you're deferring

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -182,31 +198,34 @@ func commitProcessed(ctx context.Context) {
...
-	db := datastore.GetStore().GetDB()
+	ctx = datastore.GetStore().CreateTransaction(ctx)
+	tx := datastore.GetStore().GetTransaction(ctx)
+
+	defer tx.Rollback()
+
 	var challenges []*ChallengeEntity
 
-	db.Where(ChallengeEntity{Status: Processed}).
+	tx.Where(ChallengeEntity{Status: Processed}).
```
ssardana08: reason to change it to transaction from earlier db?

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -216,28 +235,27 @@ func commitChallenge(c *ChallengeEntity) {
 			zap.String("validationTickets", string(c.ValidationTicketsString)),
 			zap.String("ObjectPath", string(c.ObjectPathString)),
 			zap.Error(err))
+		return
 	}
 
 	ctx := datastore.GetStore().CreateTransaction(context.TODO())
-	defer ctx.Done()
```
ssardana08: remove the space

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -216,28 +235,27 @@ func commitChallenge(c *ChallengeEntity) {
...
 
-	db := datastore.GetStore().GetTransaction(ctx)
+	tx := datastore.GetStore().GetTransaction(ctx)
+	var err error
+	defer func() {
```
ssardana08: use the function

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -39,19 +42,19 @@ func (cr *ChallengeEntity) SubmitChallengeToBC(ctx context.Context) (*transactio
 
 	err = txn.ExecuteSmartContract(transaction.STORAGE_CONTRACT_ADDRESS, transaction.CHALLENGE_RESPONSE, sn, 0)
 	if err != nil {
-		Logger.Info("Failed submitting challenge to the mining network", zap.String("err:", err.Error()))
+		zlogger.Logger.Info("Failed submitting challenge to the mining network", zap.String("err:", err.Error()))
```
ssardana08: we can use it as it is, no?
```go
logging.Logger.Info("Failed submitting challenge to the mining network", zap.String("err:", err.Error()))
```

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -39,19 +42,19 @@ func (cr *ChallengeEntity) SubmitChallengeToBC(ctx context.Context) (*transactio
...
 		return nil, err
 	}
 
-	Logger.Info("Verifying challenge response to blockchain.", zap.String("txn", txn.Hash), zap.String("challenge_id", cr.ChallengeID))
+	zlogger.Logger.Info("Verifying challenge response to blockchain.", zap.String("txn", txn.Hash), zap.String("challenge_id", cr.ChallengeID))
```
ssardana08: same as above

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -39,19 +42,19 @@ func (cr *ChallengeEntity) SubmitChallengeToBC(ctx context.Context) (*transactio
...
 	time.Sleep(transaction.SLEEP_FOR_TXN_CONFIRMATION * time.Second)
 
 	t, err := transaction.VerifyTransaction(txn.Hash, chain.GetServerChain())
 	if err != nil {
-		Logger.Error("Error verifying the challenge response transaction", zap.String("err:", err.Error()), zap.String("txn", txn.Hash), zap.String("challenge_id", cr.ChallengeID))
+		zlogger.Logger.Error("Error verifying the challenge response transaction", zap.String("err:", err.Error()), zap.String("txn", txn.Hash), zap.String("challenge_id", cr.ChallengeID))
```
ssardana08: same as above

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -39,19 +42,19 @@ func (cr *ChallengeEntity) SubmitChallengeToBC(ctx context.Context) (*transactio
...
 		return txn, err
 	}
-	Logger.Info("Challenge committed and accepted", zap.Any("txn.hash", t.Hash), zap.Any("txn.output", t.TransactionOutput), zap.String("challenge_id", cr.ChallengeID))
+	zlogger.Logger.Info("Challenge committed and accepted", zap.Any("txn.hash", t.Hash), zap.Any("txn.output", t.TransactionOutput), zap.String("challenge_id", cr.ChallengeID))
```
ssardana08: same as above

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -60,7 +63,7 @@ func (cr *ChallengeEntity) ErrorChallenge(ctx context.Context, err error) {
 	cr.UpdatedAt = time.Now().UTC()
 
 	if err := cr.Save(ctx); err != nil {
-		Logger.Error("[challenge]db: ", zap.String("challenge_id", cr.ChallengeID), zap.Error(err))
+		zlogger.Logger.Error("[challenge]db: ", zap.String("challenge_id", cr.ChallengeID), zap.Error(err))
```
ssardana08: logging

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -70,49 +73,57 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
 	}
 
 	allocationObj, err := allocation.GetAllocationByID(ctx, cr.AllocationID)
 	if err != nil {
 		return err
 	}
 
+	// Lock allocation changes from happening in handler.CommitWrite function
+	allocMu := lock.GetMutex(allocationObj.TableName(), allocationObj.ID)
+	allocMu.Lock()
```
ssardana08, line 96: @lpoli please add a comment on top of this function, that we should be unlocking it as soon as possible and never defer allocMu.Unlock() as it'll be concurrently required and if we defer, it'll be locked for longer time

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +182,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
-		cr.ValidationTickets[i] = &validationTicket
+		swg.Add()
+		go func(url, validatorID string, i int) {
+			defer swg.Done()
+
+			resp, err := util.SendPostRequest(url, postDataBytes, nil)
+			if err != nil {
+				zlogger.Logger.Info("[challenge]post: ", zap.Any("error", err.Error()))
+				delete(responses, validatorID)
+				cr.ValidationTickets[i] = nil
```
ssardana08: don't we need a lock here?

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +182,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+			err = json.Unmarshal(resp, &validationTicket)
+			if err != nil {
+				zlogger.Logger.Error(
+					"[challenge]resp: ",
+					zap.String("validator",
+						validatorID),
+					zap.Any("resp", string(resp)),
+					zap.Any("error", err.Error()),
+				)
+				accessMu.Lock()
```
ssardana08: create these 4 lines a function

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +182,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+				zap.Any("validator_response", validationTicket),
+			)
+			verified, err := validationTicket.VerifySign()
+			if err != nil || !verified {
+				zlogger.Logger.Error(
+					"[challenge]ticket: Validation ticket from validator could not be verified.",
+					zap.String("validator",
+						validatorID),
+				)
+				accessMu.Lock()
```
ssardana08: use a newly created function instead of repeating lines

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +182,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+				delete(responses, validatorID)
+				cr.ValidationTickets[i] = nil
+				accessMu.Unlock()
+				return
+			}
+			accessMu.Lock()
```
ssardana08: use a newly created function instead of repeating lines

ssardana08 (changes requested):

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Not sure that would be a good idea to use it as a function. 
This defer statement has condition on `err`. We need to pass `tx` and `err` as function parameter.

If `err` is updated below, it will still be `nil` and will always result in `tx.Commit()`.

One way would be to create a function:
```
func CommitOrRollBack(tx *gorm.DB, err *error){
    if err!=nil{
        tx.Rollback()
        return    
    }
    tx.Commit()
}

```
And defer function call would be:
```
var err error
defer CommitOrRollBack(tx, &err)
```

Do you suggest this instead?

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: No we don't.
Because validator can validate based on the data send by blobber.

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -169,53 +185,84 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
 	}
 	responses := make(map[string]ValidationTicket)
 	if cr.ValidationTickets == nil {
 		cr.ValidationTickets = make([]*ValidationTicket, len(cr.Validators))
 	}
+
+	accessMu := sync.Mutex{}
+	updateMapAndSlice := func(validatorID string, i int, vt *ValidationTicket) {
+		accessMu.Lock()
+		cr.ValidationTickets[i] = vt
```
peterlimg, line 205: Using temp mutex to protect the access of a member variable might not thread safe, imagine if any change the `LoadValidationTickets()` is called concurrently against the same challenge object (not sure if this could happen, but once it happens, we will panic).  So if we are sure the open challenges in https://github.com/0chain/blobber/blob/ce26910d76482b50601ffa846dd7bf50a0666488/code/go/0chain.net/blobbercore/challenge/challenge.go#L127 would never duplicate, then perhaps we are safe.

peterlimg (dismissed):

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: How will it panic?
It is totally in the scope of the function.

Even if it is called concurrently, it will initialize its own mutex. Don't understand how it will panic.
@peterlimg

code/go/0chain.net/blobbercore/challenge/challenge.go
ssardana08: yeah we can pass err as reference
duplicating same code everywhere for txs doesn't make sense to me

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -70,49 +73,60 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
 
+		logging.Logger.Debug(cr.StatusMessage + "for challenge: " + cr.ChallengeID)
+
 		if err := cr.Save(ctx); err != nil {
-			Logger.Error("[challenge]db: ", zap.String("challenge_id", cr.ChallengeID), zap.Error(err))
+			logging.Logger.Error("[challenge]db: ", zap.String("challenge_id", cr.ChallengeID), zap.Error(err))
+			return err
 		}
-		return common.NewError("no_validators", "No validators assigned to the challenge")
+		return nil
```
cnlangzi, line 83: why the error is removed? it is invalid challenge if validators is empty.

cnlangzi (commented):
-

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: It was supposed to be stored in database.
returning error will prevent it from being stored in database

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -70,49 +73,60 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
 
 	rootRef, err := reference.GetReference(ctx, cr.AllocationID, "/")
+	if err != nil {
+		allocMu.Unlock()
+		cr.ErrorChallenge(ctx, err)
+		return err
+	}
+
 	blockNum := int64(0)
 	if rootRef.NumBlocks > 0 {
```
ssardana08, line 116: in numBlocks == 0, shouldn't we stop process here?

ssardana08 (changes requested):

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: No.
This is the actual fix for blank allocation.

Even if there is not files in allocation, it should still provide validators with list of writemarkers so that validator can verify the challenge.

code/go/0chain.net/blobbercore/challenge/protocol.go
ssardana08: @peterlimg chances of same chalenge_id coming many times might be there to `validateChallenge`, but all those would be different object. will that cause panic?

ssardana08 (approved):
lgtm

code/go/0chain.net/blobbercore/challenge/protocol.go
cnlangzi: hmm, the old logic seems be a problem. maybe `status` and `result` should also be updated with `Cancelled` and `ChallengeFailure`? if not, it still is processed in next timer trigger. And we can't get any error from logging about this error now.  @lpoli   do you think it is better? if you think it is not better, could we also print this error message in logs?

PR #776, opened 2022-08-02
fix(challenge): add time stopwatch for workers

cnlangzi:
### Changes
- added time stopwatch for challenge workers.

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/blobbercore/challenge/map.go
```diff
@@ -0,0 +1,45 @@
+package challenge
+
+import (
+	"sync"
+)
+
+var cMap = &ChallengeMap{
+	items: make(map[string]ChallengeStatus),
```
peterlimg: Use the LRU cache to avoid potential OOM, https://github.com/0chain/blobber/blob/6041fff4855a9ffee008b4bf8ec5e6ff5b7b3caa/code/go/0chain.net/core/cache/lru.go#L13.

We should be able to use it directly to replace the ChallengeMap

peterlimg (changes requested):

code/go/0chain.net/blobbercore/challenge/map.go
cnlangzi: fixed

peterlimg:
I mean the LRUCache is thread safe itself, we can use it directly, so we don't need the `map.go`

cnlangzi:
> I mean the LRUCache is thread safe itself, we can use it directly, so we don't need the `map.go`

fixed

peterlimg (approved):
lgtm

PR #783, opened 2022-08-05
Avoid processing never verifiable challenge

lpoli:
### Fixes
This PR will avoid processing challenges that will fail eventually in Blockchain.

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/blobbercore/challenge/challenge.go
+++ b/code/go/0chain.net/blobbercore/challenge/challenge.go
@@ -141,7 +141,9 @@ func processAccepted(ctx context.Context) {
 
 	rows, err := db.Model(&ChallengeEntity{}).
 		Where("status = ?", Accepted).
-		Select("challenge_id", "created_at").Rows()
+		Select("challenge_id", "created_at").
+		Order("sequence").
+		Rows()
 
 	if err != nil {
 		logging.Logger.Error("[challenge]process: ",
@@ -286,7 +288,9 @@ func commitProcessed(ctx context.Context) {
 
 	rows, err := db.Model(&ChallengeEntity{}).
 		Where("status = ?", Processed).
-		Select("challenge_id", "created_at").Rows()
+		Select("challenge_id", "created_at").
+		Order("sequence").
+		Rows()
 
 	if err != nil {
 		logging.Logger.Error("[challenge]commit: ",
--- a/code/go/0chain.net/blobbercore/challenge/protocol.go
+++ b/code/go/0chain.net/blobbercore/challenge/protocol.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"math"
 	"math/rand"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/allocation"
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/filestore"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/reference"
 	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/writemarker"
@@ -24,6 +26,7 @@ import (
 )
 
 const VALIDATOR_URL = "/v1/storage/challenge/new"
+const ValueNotPresent = "value not present"
 
 type ChallengeResponse struct {
 	ChallengeID       string              `json:"challenge_id"`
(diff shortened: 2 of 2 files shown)


code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -141,7 +141,9 @@ func processAccepted(ctx context.Context) {
 
 	rows, err := db.Model(&ChallengeEntity{}).
 		Where("status = ?", Accepted).
-		Select("challenge_id", "created_at").Rows()
+		Select("challenge_id", "created_at").
+		Order("sequence").
```
ssardana08, line 145: we should maintain last processed sequence per each process and make sure when querying we select `sequence > last_processed_sequence` of that process
this is to ensure not same challenge id is processed again

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -286,7 +288,9 @@ func commitProcessed(ctx context.Context) {
 
 	rows, err := db.Model(&ChallengeEntity{}).
 		Where("status = ?", Processed).
-		Select("challenge_id", "created_at").Rows()
+		Select("challenge_id", "created_at").
+		Order("sequence").
```
ssardana08, line 292: same as above
something to ensure no same challenge is taken again

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -301,6 +304,11 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
 }
 
 func (cr *ChallengeEntity) CommitChallenge(ctx context.Context, verifyOnly bool) error {
+	if time.Since(cr.CreatedAt) > config.Configuration.ChallengeCompletionTime+time.Second {
```
ssardana08, line 307: why `+time.Second`?

ssardana08, line 307: @cnlangzi @lpoli we should not set cct in config
but get from this API `sharder01/v1/screst/6dba10422e368813802877a85039d3985d96760ed844092319743fb3a76712d7/storage-config` key: `max_challenge_completion_time`

ssardana08: if this is for blobber server delay.. that's quite unpredictable
we don't know if it's ahead or behind..

ssardana08: we need to ensure the times are same

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -301,6 +304,11 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+		cr.Status = Cancelled
+		cr.StatusMessage = "challenge completion time expired"
+		cr.UpdatedAt = time.Now().UTC()
+	}
```
ssardana08, line 311: why do we need to submitChallenge if we hit this scenario?

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -337,6 +345,12 @@ func (cr *ChallengeEntity) CommitChallenge(ctx context.Context, verifyOnly bool)
 			cr.LastCommitTxnIDs = append(cr.LastCommitTxnIDs, t.Hash)
 			cr.UpdatedAt = time.Now().UTC()
 		}
+
+		if IsValueNotPresentError(err) {
```
ssardana08, line 349: cr.UpdatedAt should be updated

ssardana08 (changes requested):

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: @sum-it08 Same challenge id is never stored in the first place.

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Challenge can be processed but its transaction may not be verified at that moment.
Later on it will be verified. So to update its status as `committed` it can run. It will not send challenge commit transaction to the blockchain.

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: It was missed type.
I have deleted it in another PR

lpoli: Lets make another PR.
For cct, blobber should call sharder once and store this value in some unexported variable exposed with exported function

lpoli:
Closing this PR because its all changes and recently updated code is in https://github.com/0chain/blobber/pull/784

PR #802, opened 2022-08-17
Custom nonce monitor for blobber txns

avanaur:
### Changes

Use a custom nonce monitor
- When nonce is currently used/assigned to a txn, it is reserved and cannot be used on next txn until verified or reservation has lapsed
- Reset only nonce from balance when consistently failing thus avoidig

This change yields higher success rate on blobber txns which is executed from one wallet through gosdk. The change is primarily to avoid use of same nonce on inflight txns. On gosdk cacheing, this is not guaranteed as each txn verification failure results to cache evict and when happening concurrently results to same nonce being used.

image

### Fixes

Missing challenges on allocations created later on the life of the network.

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
looks good, we should keep in mind, that this aproach will reorder transactions, and transactions we assume to be executed in the order will be executed out of order

PR #803, opened 2022-08-17
Add malicious blobber behaviours

fabioDMFerreira:
Malicious blobber:
* sends wrong data on listing
* sends wrong metadata on listing
* does not respond on listing/deleting/uploading/downloading
* returns an error on listing/deleting/uploading/downloading


code/go/0chain.net/validatorcore/storage/handler_integration_tests.go
```diff
@@ -0,0 +1,13 @@
+//go:build integration_tests
```
dabasov, line 1: looks like this file is the same as handler_main.gp

dabasov (approved):
lgtm

code/go/0chain.net/validatorcore/storage/handler_integration_tests.go
fabioDMFerreira: It is. This is part of validator conductor tests development. We'll need to override the endpoint handler `/v1/storage/challenge/new`.

PR #804, opened 2022-08-18
Fix/rate limiter

lpoli:
### Changes
This PR modifies the use of tollbooth's ratelimiter and corrects use of ratelimiter with proxy on.

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/core/common/admin.go
```diff
@@ -0,0 +1,31 @@
+package common
+
+import (
+	"net/http"
+
+	"github.com/spf13/viper"
+)
+
+var username, password string
```
peterlimg: Perhaps rename to gUsername and gPassword to different them from local variables and make it obvious for users to know that they are global variables, otherwise they could be overwritten mistakenly.

code/go/0chain.net/core/common/admin.go
```diff
@@ -0,0 +1,31 @@
...
+	username = viper.GetString("admin.username")
+	password = viper.GetString("admin.password")
+}
+
+func AuthenticateAdmin(handler ReqRespHandlerf) ReqRespHandlerf {
+	return func(w http.ResponseWriter, r *http.Request) {
+		uname, passwd, ok := r.BasicAuth()
+		if !ok {
+			w.WriteHeader(http.StatusForbidden)
+			w.Write([]byte("Admin only api")) // nolint
```
peterlimg, line 22: We should return instead of continue

code/go/0chain.net/core/common/admin.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+
+		if uname != username || passwd != password {
+			w.WriteHeader(http.StatusForbidden)
+			w.Write([]byte("Invalid username or password")) // nolint
```
peterlimg, line 28: Return

code/go/0chain.net/core/common/rate_limiter.go
```diff
@@ -1,159 +1,162 @@
...
-	if fileRateLimit == nil || !fileRateLimit.RateLimit {
-		return handler
+	if gRps > 0 {
+		generalRL.SetMax(gRps)
 	}
 
-	lmt := fileRateLimit.Limiter
+}
 
+func RateLimit(handler ReqRespHandlerf, rlType RPS) ReqRespHandlerf {
```
peterlimg: Instead of using the `rlType` to different different rate limit and expose those types to users, I would suggest to define different rate limit functions and users can call them directly, they don't need to know the `RPS`, all of these could be private. For example, we can define

```
func RateLimitByCommit(handle ReqRespHandlerf) ReqRespHandlerf {
    return rateLimit(handler, LImitByCommitRL)
}

func RateLimitByFile(handler ReqRespHandlerf) ReqRespHandlerf {
    return rateLimit(handler, LimitByFileRL)
}
...

```

Then in the API handlers, we can use it like:

```go

r.HandleFunc("/v1/file/upload/{allocation}",
		common.RateLimitByFile(common.ToJSONResponse(WithConnection(UploadHandler))))

and
r.HandleFunc("/v1/connection/commit/{allocation}",
common.RateLimitByCommit(common.ToStatusCode(WithStatusConnection(CommitHandler))))

```

code/go/0chain.net/validatorcore/storage/handler.go
```diff
@@ -2,23 +2,95 @@ package storage
...
+
+func init() {
+	defaultIPLookups := []string{"RemoteAddr", "X-Forwarded-For", "X-Real-IP"}
+	lmt = tollbooth.NewLimiter(RequestPerSecond, &limiter.ExpirableOptions{
+		DefaultExpirationTTL: DefualtExpirationTime,
+	})
+
+	lmt.SetIgnoreURL(true)
+	lmt.SetIPLookups(defaultIPLookups)
+}
```
peterlimg: The rate limit code in this file seems duplicate the work that we've done in common package. I can see why this is needed, but I would suggest to define functions in common to new single rate limit with configs passed in as param instead of reading from viper directly. In that way, we can reuse the rate limit code in common package.

Maybe we should also avoid global `limits` variables defined in common package, those variables should be created by the users with `NewRateLimit()` function and stored in their top level main packages and to be reused.

peterlimg (changes requested):

code/go/0chain.net/core/common/rate_limiter.go
```diff
@@ -1,159 +1,89 @@
...
-	}
-
-	// OR add path (may exists for download)
-	path := r.Header.Get("X-Path")
-	if path != "" {
-		baseKeys[0] = append(baseKeys[0], path)
-		return baseKeys[0]
+func RateLimitByIP(handler ReqRespHandlerf, lmt *limiter.Limiter) ReqRespHandlerf {
+	return func(w http.ResponseWriter, r *http.Request) {
+		TryParseForm(r)
```
dabasov: why do we need it?

code/go/0chain.net/core/common/rate_limiter.go
```diff
@@ -1,159 +1,89 @@
...
+
+		if lmt != nil {
+
+			keys := tollbooth.BuildKeys(lmt, r)
+			for _, k := range keys {
+				httpError := tollbooth.LimitByKeys(lmt, k)
```
dabasov, line 60: why do we need to limit for every key here, it is all headers is it really needed?

dabasov (dismissed):

code/go/0chain.net/core/common/rate_limiter.go
lpoli: Which one do you mean?
`RateLimitByIP`?

`RateLimitByIP` will limit request based only on client's ip. It is used by validator.

code/go/0chain.net/core/common/rate_limiter.go
lpoli: Keys is of type `[][]string`.
`tollbooth.BuildKeys(lmt, r)` will return `[[keys extracted from request headers...]]` and its length is 1. So it is single iteration only.

It is equivalent of doing, `httpError := tollbooth.LimitByKeys(lmt, keys[0])`

code/go/0chain.net/core/common/rate_limiter.go
dabasov: I mean ``TryParseForm(r)`` call in the rate limiter

code/go/0chain.net/core/common/rate_limiter.go
dabasov: I can't get what we are doing here, we are computing key of the bucket by concatenating all the headers, what is it for, should we use only one header here, like ip or forwarded-ip?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/upload-issue  |
| gosdk  | fix/upload-issue  |
| zboxcli  | fix/upload-issue  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/rate-limiter  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -436,10 +444,17 @@ func (fsh *StorageHandler) CommitWrite(ctx context.Context, r *http.Request) (*b
...
+	elapsedWritePreRedeem := time.Since(startTime) - elapsedAllocation - elapsedGetLock -
+		elapsedGetConnObj - elapsedVerifyWM
+
 	err = connectionObj.ApplyChanges(ctx, writeMarker.AllocationRoot, writeMarker.Timestamp)
 	if err != nil {
 		return nil, err
 	}
+
+	elapsedApplyChanges := time.Since(startTime) - elapsedAllocation - elapsedGetLock -
+		elapsedGetConnObj - elapsedVerifyWM - elapsedWritePreRedeem
```
peterlimg, line 456: Re-declaring the startTime or a new timestamp will make the elapsed* calculation much simpler. So the elapsedApplyChanges would be simple be `time.Since(startTime)`.  So as all the elapsed* above.

code/go/0chain.net/core/common/rate_limiter.go
```diff
@@ -1,159 +1,86 @@
...
-//UserRateLimit - rate limiting for end user handlers
-func UserRateLimit(handler ReqRespHandlerf) ReqRespHandlerf {
-	if !userRateLimit.RateLimit {
-		return handler
-	}
-	return func(writer http.ResponseWriter, request *http.Request) {
-		tollbooth.LimitFuncHandler(userRateLimit.Limiter, handler).ServeHTTP(writer, request)
-	}
-}
+			keys = append(keys, []string{clientID})
```
peterlimg, line 35: Check if clientID is empty before append it to keys.

peterlimg (changes requested):

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: This will be deleted once we are over with loadtest.
However I will make the change for simplicity

code/go/0chain.net/core/common/rate_limiter.go
lpoli: I think we should also have empty client key.
We must mandate presence of clientID in the header.
Only some endpoints don't have clientID requirement. But we should verify clientID as well later on. So empty clientID as a key in rate-limiter will prevent un-necessary access.

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/rate-limiter  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/rate-limiter  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/rate-limiter  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | fix/rate-limiter  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #821, opened 2022-09-09
Set/Update cct from blockchain.

### LINKED ISSUE #818: Remove challenge completion time.
Challenge completion time is no longer determined on a per blobber basis, so we can remove all reference to challenge completion time. from the blobber; Including the [config.yaml setting](https://github.com/0chain/blobber/blob/staging/config/0chain_blobber.yaml#L70).

This is related to https://github.com/0chain/0chain/issues/1639.

lpoli:
### Changes
This PR will change setting of cct value from config file to api leveraged by blockchain.
Will fix https://github.com/0chain/blobber/issues/818

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/blobber/config.go
+++ b/code/go/0chain.net/blobber/config.go
@@ -134,7 +134,10 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.Geolocation.Latitude = viper.GetFloat64("geolocation.latitude")
 	config.Configuration.Geolocation.Longitude = viper.GetFloat64("geolocation.longitude")
 
-	config.Configuration.ChallengeCompletionTime = viper.GetDuration("challenge_completion_time")
+	err := setCCTFromChain()
+	if err != nil {
+		panic(err)
+	}
 
 	fmt.Print("		[OK]\n")
 }
--- a/code/go/0chain.net/blobber/config.go
+++ b/code/go/0chain.net/blobber/config.go
@@ -134,11 +134,6 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.Geolocation.Latitude = viper.GetFloat64("geolocation.latitude")
 	config.Configuration.Geolocation.Longitude = viper.GetFloat64("geolocation.longitude")
 
-	err := setCCTFromChain()
-	if err != nil {
-		panic(err)
-	}
-
 	fmt.Print("		[OK]\n")
 }
 
--- a/code/go/0chain.net/blobber/settings.go
+++ b/code/go/0chain.net/blobber/settings.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"time"
+
+	"github.com/0chain/blobber/code/go/0chain.net/blobbercore/config"
+	"github.com/0chain/blobber/code/go/0chain.net/core/logging"
+	"github.com/0chain/gosdk/zcncore"
+)
+
+type cctCB struct {
+	done chan struct{}
+	cct  time.Duration
+	err  error
+}
+
+func (c *cctCB) OnInfoAvailable(op int, status int, info string, errStr string) {
+	defer func() {
+		c.done <- struct{}{}
+	}()
+
+	if errStr != "" {
+		c.err = errors.New(errStr)
+		return
+	}
+
+	m := make(map[string]interface{})
+	err := json.Unmarshal([]byte(info), &m)
+	if err != nil {
+		c.err = err
+		return
+	}
+
+	m = m["fields"].(map[string]interface{})
+	cct := m["max_challenge_completion_time"].(string)
+
+	d, err := time.ParseDuration(cct)
+	if err != nil {
+		c.err = err
+		return
+	}
+	c.cct = d
+}
+
+func setCCTFromChain() error {
+	cb := &cctCB{
+		done: make(chan struct{}),
+	}
+	err := zcncore.GetStorageSCConfig(cb)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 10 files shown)


code/go/0chain.net/blobber/config.go
```diff
@@ -134,7 +134,10 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.Geolocation.Latitude = viper.GetFloat64("geolocation.latitude")
 	config.Configuration.Geolocation.Longitude = viper.GetFloat64("geolocation.longitude")
 
-	config.Configuration.ChallengeCompletionTime = viper.GetDuration("challenge_completion_time")
+	err := setCCTFromChain()
+	if err != nil {
+		panic(err)
```
Sriep: Why use a panic here? If I understand, an error here indicates an issue with 0chain rather than the blobber. 

Maybe the problem here is that the challenge completion time does not belong in the configuration file anymore on account of being removed, and we should move the challenge completion time elsewhere. 

https://github.com/0chain/blobber/blob/staging/code/go/0chain.net/blobber/main.go#L12 Looking at `blobber/main.go`, the first thing the blobber does is read the config file and use it to set the config object; Setting the challenge completion time in setupConfig seems out of place. Can we guarantee 0chain is reachable at this point?

code/go/0chain.net/blobbercore/config/config.go
```diff
@@ -171,7 +170,7 @@ type Config struct {
 	AutomaticUpdate       bool
 	BlobberUpdateInterval time.Duration
 
-	ChallengeCompletionTime time.Duration
+	CCT time.Duration // challenge completion time // retrieved from storage SC config
```
Sriep: Why the change to CCT? ChallengeCompletionTime looks clearer to me. 

You can remove the extra `//`.

code/go/0chain.net/blobber/settings.go
```diff
@@ -0,0 +1,85 @@
...
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			// We'd panic if err occurred when calling setCCTFromChain from
+			// main.go file because cct would be initially 0 and we cannot
+			// work with 0 value.
+			// Upon updating cct, we only log error because cct is not 0
+			// We should try to submit challenge as soon as possible regardless
+			// of cct value.
+			err := setCCTFromChain()
```
Sriep, line 79: If we set the challenge completion time here, why do we need to set it in setupConfig where it does not seem to belong?

Sriep (changes requested):
Maybe get someone more familiar with the blobber to suggest the best place to set the challenge completion time.

code/go/0chain.net/blobber/config.go
lpoli: Though it is not blobber error but of 0chain, we cannot start blobber with 0 cct.
What is better here?

Also the value is retrieved from storage-config endpoint. It is also termed as storageSCconfig and isn't it better to put in configuration file?

Don't understand what you mean by `0chain is reachable at this point`. Basically it is only asking for storageSCConfig which is publicly available without wallet.

code/go/0chain.net/blobber/settings.go
lpoli: Here it is done to update CCT. It can have been changed dynamically (which is not the current case though).

Do you think it can be dynamic? Or will it be dead constant?

Sriep: The challenge completion time can change with an update settings transaction. https://github.com/0chain/zwalletcli/blob/staging/cmd/sc-update-config.go#L14
So the blobber needs to be able to respond.

code/go/0chain.net/blobber/config.go
Sriep: The challenge competition time is no longer in the configuration file, so it smells bad to keep it in the configuration file's model.  And setting it in the method designed to read in the configuration file smells just as bad. Better create a sister method to [setupConfig](https://github.com/0chain/blobber/blob/staging/code/go/0chain.net/blobber/main.go#L12) to get settings from 0chain.

If there is no object for settings read from 0chain yet then we should create one. We could create a parent object for `config.Configuration`, and have 0chain settings as a sister. Anyway, check with someone more familiar with the blobber.

code/go/0chain.net/blobber/config.go
```diff
@@ -134,7 +134,10 @@ func setupConfig(configDir string, deploymentMode int) {
 	config.Configuration.Geolocation.Latitude = viper.GetFloat64("geolocation.latitude")
 	config.Configuration.Geolocation.Longitude = viper.GetFloat64("geolocation.longitude")
 
-	config.Configuration.ChallengeCompletionTime = viper.GetDuration("challenge_completion_time")
+	err := setCCTFromChain()
```
Sriep: Can we move this outside of setupConfig, this has nothing to do with setting up the config file. Maybe have the main function call setupSettings that calls setupConfig and setCCTFromChain? Or alternately add a setupSettingsFrom0Chain method to the main funtion.

Sriep (changes requested):

Sriep (approved):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #859, opened 2022-10-22
Fast fail txns

Kishan-Dhakan:
### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


cnlangzi (commented):
LGTM. please fix unit tests. thanks

code/go/0chain.net/blobbercore/allocation/protocol.go
```diff
@@ -50,6 +50,12 @@ func (a *Allocation) LoadTerms(ctx context.Context) (err error) {
 
 // VerifyAllocationTransaction try to get allocation from postgres.if it doesn't exists, get it from sharders, and insert it into postgres.
 func VerifyAllocationTransaction(ctx context.Context, allocationTx string, readonly bool) (a *Allocation, err error) {
+	t, err := transaction.VerifyTransaction(allocationTx, chain.GetServerChain())
```
lpoli: Actually Previous logic is better than this because if the allocation transaction `allocationTx`( which is different from `allocationID`) is same as it is in the database then we don't need to verify the transaction.

Transaction verification is costlier than retrieving from database. So Please revert it back.

Note; `allocationID` is the `allocationTx` in initial allocation creation time. Later when allocation is updated then these two differ.

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -370,18 +370,20 @@ func (fsh *StorageHandler) RemoveCollaborator(ctx context.Context, r *http.Reque
 
 func (fsh *StorageHandler) GetFileStats(ctx context.Context, r *http.Request) (interface{}, error) {
 	allocationTx := ctx.Value(constants.ContextKeyAllocation).(string)
+	clientSign, _ := ctx.Value(constants.ContextKeyClientSignatureHeaderKey).(string)
+	clientPublicKey, _ := ctx.Value(constants.ContextKeyClientKey).(string)
```
lpoli, line 374: If you are using `clientPublicKey` from http header then it means `GetFileStats` is public endpoint and a subject to ddos attack.

If it is meant to be to only the owner of an allocation then we must use `allocationObj.OwnerPublicKey` in signature verification.

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -627,18 +629,19 @@ func (fsh *StorageHandler) getReferencePath(ctx context.Context, r *http.Request
 
 func (fsh *StorageHandler) GetObjectPath(ctx context.Context, r *http.Request) (*blobberhttp.ObjectPathResult, error) {
 	allocationTx := ctx.Value(constants.ContextKeyAllocation).(string)
+	clientSign, _ := ctx.Value(constants.ContextKeyClientSignatureHeaderKey).(string)
+	ClientPublicKey, _ := ctx.Value(constants.ContextKeyClientKey).(string)
+	valid, err := verifySignatureFromRequest(allocationTx, clientSign, ClientPublicKey)
```
lpoli: Same as above. Also I think we can remove `GetObjectPath` endpoint. It is not required anywhere.

lpoli (changes requested):

code/go/0chain.net/blobbercore/handler/storage_handler.go
stewartie4: Kishan is out for a week @lpoli, are you ok to make these changes to keep things moving?

boddumanohar: @stewartie4  I will be addressing the PR comments. Thanks.

boddumanohar: created a seperate PR for this. https://github.com/0chain/blobber/pull/863

code/go/0chain.net/blobbercore/allocation/protocol.go
Kishan-Dhakan: Okay, will revert.

code/go/0chain.net/blobbercore/handler/storage_handler.go
Kishan-Dhakan: Actually GetFileStats is a public endpoint whether I clientPublicKey from header or from allocationObj. It is a subject to DDoS attack in both cases, as we have seen in our DDoS tests. The difference is that getting it from allocationObj takes 60000 ms to respond (median response time), while getting it from header to fail the txns takes only ~100 ms (median response time). Moreover, security is also not an issue as whether we get it from header or we get it from allocationObj, we call verifySignatureFromRequest so we know if it's incorrect.

boddumanohar:
@Kishan-Dhakan  is this PR still relevant?

Kishan-Dhakan:
@boddumanohar closing this, will make a new PR with the changes.

PR #865, opened 2022-10-27
stop checking for hash

### LINKED ISSUE #860: too less sharders to confirm
in this piece of code, we checking if the response JSON returned from all the sharders is the same by check for the hash of the responses

https://github.com/0chain/blobber/blob/8dc6681f7795bee95eeef8c56ea09ce2d7ceffdf/code/go/0chain.net/core/transaction/http.go#L137-L152

But sometimes even if the responses of the JSON is the same, the order of items in the JSON are different, that case even if the JSON objects are equivalent, the blobber still thinks that the information returned is in-correct since we are checking for hashes.

Screenshot 2022-10-22 at 13 37 56

boddumanohar:
### Changes
Instead of checking for hash of the response, Just check if the response is received or not.

### Fixes
fixes: https://github.com/0chain/blobber/issues/860

### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/core/transaction/http.go
```diff
@@ -143,13 +142,17 @@ func makeSCRestAPICall(scAddress string, relativePath string, params map[string]
 			return errors.Throw(ErrBadRequest, errorMsg)
 		}
 
-		hashString := hex.EncodeToString(hash.Sum(nil))
-		hashCounters[hashString]++
+		resMaxCounterBody = resBody
```
peterlimg, line 149: This would only return the last response, and no consensus is actually met. This can be a workaround for the fix. But actually it's hard to have consensus as the sharders could be in different LFB when they receive the request, which means they would give different response.

code/go/0chain.net/core/transaction/http.go
```diff
@@ -143,13 +142,17 @@ func makeSCRestAPICall(scAddress string, relativePath string, params map[string]
...
-			resMaxCounterBody = resBody
-		}
+		// TODO: Instead of the hash, we should JSON Equivance
+		// hashString := hex.EncodeToString(hash.Sum(nil))
+		// hashCounters[hashString]++
+
+		// if hashCounters[hashString] > hashMaxCounter {
+		// 	hashMaxCounter = hashCounters[hashString]
+		// 	resMaxCounterBody = resBody
+		// }
```
peterlimg: Remove the code if not used

code/go/0chain.net/core/transaction/http.go
```diff
@@ -71,7 +70,7 @@ func makeSCRestAPICall(scAddress string, relativePath string, params map[string]
 	var resMaxCounterBody []byte
 
 	var hashMaxCounter int
-	hashCounters := make(map[string]int)
+	// hashCounters := make(map[string]int)
```
peterlimg: remove if not used

peterlimg (changes requested):

code/go/0chain.net/core/transaction/http.go
boddumanohar: removed

code/go/0chain.net/core/transaction/http.go
boddumanohar: removed.

peterlimg (approved):

code/go/0chain.net/core/transaction/http.go
boddumanohar: added this PR comment as a `note` in the code.

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | too-few-sharders  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1007, opened 2023-03-22
Fix/challenge issues parallel

lpoli:
### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -1,408 +1,162 @@
...
-	}
-
-	if err := tx.Commit().Error; err != nil {
-		logging.Logger.Error("[challenge]validate(Commit): ",
-			zap.Any("challenge_id", c.ChallengeID),
-			zap.Time("created", createdTime),
-			zap.Error(err))
-		tx.Rollback()
-		return
+		// offset += len(challengeResponse.Challenges)
```
dabasov, line 97: this query won't work without offset, every time you request you will get the same challenge, so you will duplicate the work

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -1,408 +1,162 @@
...
 	for {
-		var challenges BCChallengeResponse
-		var challengeIDs []string
-		challenges.Challenges = make([]*ChallengeEntity, 0)
-		apiStart := time.Now()
-		retBytes, err := transaction.MakeSCRestAPICall(transaction.STORAGE_CONTRACT_ADDRESS, "/openchallenges", params, chain.GetServerChain())
-		if err != nil {
-			logging.Logger.Error("[challenge]open: ", zap.Error(err))
-			break
+		d := time.Since(challengeRespTime)
```
dabasov, line 48: handle context to break the loop

code/go/0chain.net/blobbercore/challenge/challenge.go
```diff
@@ -1,408 +1,162 @@
...
+		guideCh <- struct{}{}
+
+		seqManagerCh <- chalEntity
+		go func(chalEntity *ChallengeEntity) {
+			defer func() {
+				<-guideCh
+			}()
+			logging.Logger.Info("Processing challenge", zap.String("challenge_id", chalEntity.ChallengeID))
+			ctx := datastore.GetStore().CreateTransaction(context.TODO())
+			defer func() {
```
dabasov, line 119: don't use several defer blocks

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -354,111 +287,284 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+					oldestCresp = oldestCresp.nextCResp
+				} else {
+					break
 				}
+			}
 
-				return nil
+			// deplete all the challenges in buffer
+		D:
+			for {
```
dabasov, line 406: this will cause spin wait, avoid it by all means

code/go/0chain.net/blobbercore/challenge/protocol.go
```diff
@@ -354,111 +287,284 @@ func (cr *ChallengeEntity) LoadValidationTickets(ctx context.Context) error {
...
+// C1 is still processing.
+// C3 is not going to work because its previous nonce is missing.
+// First we should check if C1 is still processing and if so wait for it.
+// Otherwise if we send C3 with nonce 2 and then C1 fails then again C3 will fail.
+// Wait for C1. If comes with error then remove it and send C3 with nonce 1.
+//
+// We need to also take care of the fact that there are other processes(write/read marker redeeming) that uses
+// nonceMonitor which will update latest nonce value.
+
+func ProcessChallengeTransactions(ctx context.Context) {
```
dabasov, line 322: this func should be rewritten

dabasov (changes requested):

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Wouldn't `from` param exactly filter it? Otherwise what use is of `from` param?
If we use both `from` and `offset` then it seems problem to me.

code/go/0chain.net/blobbercore/challenge/challenge.go
lpoli: Why so?
There are only two defer blocks.

code/go/0chain.net/blobbercore/challenge/protocol.go
lpoli: No it won't. The loop is protected by lock

dabasov: how it can help? you will enter the loop go to default and enter it again. we should never use spin waits

PR #1037, opened 2023-04-21
Enable blobber to upload multiple file

Roshan-Mehta:
### Changes
 1. The blobber didn't support multiple file uploads with a single connectionID; it was creating a single entry in the AllocationChange table. As part of this PR, we have checked if the chunk coming is of which file and then append it accordingly. 
 2. We are completely removing the lock from object_handler. We are tracking connectionObjSize in memory.
### Tests

- 0chain:
- gosdk: https://github.com/0chain/gosdk/pull/909
- system_test:
- zboxcli: https://github.com/0chain/zboxcli/pull/402
- zwalletcli:
- Other: ...

Benchmark result
Screenshot 2023-05-08 at 9 07 25 PM


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | feature/multi_operation  |
| zboxcli  | feature/parallelizing-sync-command  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | feature/multi_operation  |
| zboxcli  | feature/parallelizing-sync-command  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | feature/multi_operation  |
| zboxcli  | feature/parallelizing-sync-copy  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

code/go/0chain.net/core/lock/lock.go
```diff
@@ -37,6 +38,21 @@ func (m *Mutex) Unlock() {
 	m.mu.Unlock()
 }
 
+// TryUnlock unlock the lock if it is already locked otherwise do nothing.
+// Don't use tryUnlock and Unlock together for the same lock otherwise it may panic
+func (m *Mutex) TryUnlock() {
+	m.tryUnlockMu.Lock()
```
dabasov: why do you need second mutex?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -1041,24 +1038,19 @@ func (fsh *StorageHandler) WriteFile(ctx context.Context, r *http.Request) (*blo
...
 	Logger.Info(fmt.Sprintf("[upload] Acquiring lock for allocation: %s, connection: %s", allocationID, connectionID))
-	mutex := lock.GetMutex(connectionObj.TableName(), connectionID)
+	mutex := lock.GetMutex(AllocationConnections, connectionID) // It will be unlocked from WithConnection();
 
 	Logger.Info(fmt.Sprintf("[upload] Locking for allocation: %s, connection: %s", allocationID, connectionID))
 
 	mutex.Lock()
 	Logger.Info(fmt.Sprintf("[upload] Acquired lock for allocation: %s, connection: %s", allocationID, connectionID))
 
-	defer func() {
```
dabasov, line 1057: why is unlocking block removed?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -947,15 +946,13 @@ func (fsh *StorageHandler) CreateDir(ctx context.Context, r *http.Request) (*blo
 		return nil, common.NewError("invalid_parameters", "Invalid connection id passed")
 	}
 
+	mutex := lock.GetMutex(AllocationConnections, connectionID) // It will be unlocked from WithConnection();
+	mutex.Lock()
```
dabasov: why no defer Unlock?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -780,13 +780,12 @@ func (fsh *StorageHandler) MoveObject(ctx context.Context, r *http.Request) (int
...
```
dabasov: why no defer unlock?

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -671,13 +671,13 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
 		return nil, common.NewError("invalid_parameters", "Invalid connection id passed")
 	}
 
+	mutex := lock.GetMutex(AllocationConnections, connectionID) // It will be unlocked from WithConnection();
+
+	mutex.Lock()
```
dabasov: why no defer unlock

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -581,15 +582,14 @@ func (fsh *StorageHandler) RenameObject(ctx context.Context, r *http.Request) (i
 		return nil, common.NewError("invalid_parameters", "Invalid connection id passed")
 	}
 
+	mutex := lock.GetMutex(AllocationConnections, connectionID) // It will be unlocked from WithConnection();
+	mutex.Lock()
```
dabasov: hy no defer unlock?

dabasov (changes requested):

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: We shouldn't. 
It will create a problem if we unlock it here. 
The flow is like that, allocation.GetAllocationChanges() function returns connectionObj. It marks the connection as a new connection if the connectionId is not present in the AllocationChange table. 
This function `RenameObject`, for example, doesn't "actually" save the record in DB. The record is saved in WithConection() method. So if unlock the lock in the `RenameObject` function, then there might be the case that before the record is saved in the DB, the other operation runs (with old connectionId) arrive. This connection is again marked as a new connection!
And this was happening during my local testing. I discussed this with @lpoli about this issue, and we arrived at this approach.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: https://github.com/0chain/blobber/pull/1037#discussion_r1182907578

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: https://github.com/0chain/blobber/pull/1037#discussion_r1182907578

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: https://github.com/0chain/blobber/pull/1037#discussion_r1182907578

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: https://github.com/0chain/blobber/pull/1037#discussion_r1182907578

code/go/0chain.net/core/lock/lock.go
Roshan-Mehta: I can't use lockMutex because I am calling the unlock method after locking the mutex, and that unlock method uses lockMutex. So, it will be infinite waiting.

Hitenjain14:
Why do we need to update the `connectionObj.Size` after every operation? When client commits the data, calculate the size by iterating over the changes, or is there any other place where `connectionObj.Size` is getting used

Roshan-Mehta:
> Why do we need to update the `connectionObj.Size` after every operation? When client commits the data, calculate the size by iterating over the changes, or is there any other place where `connectionObj.Size` is getting used

@Hitenjain14  It is used during chunk upload to check if the size limit is exceeded for this blobber.

Hitenjain14:
> > Why do we need to update the `connectionObj.Size` after every operation? When client commits the data, calculate the size by iterating over the changes, or is there any other place where `connectionObj.Size` is getting used
> 
> @Hitenjain14 It is used during chunk upload to check if the size limit is exceeded for this blobber.

Got your point, one way I can think of to get rid of this locking is by  using the size of tempDir of that connectionID to find out if client is going above the limit. There can be an issue with thumbnail files as we upload the content with it too, but if we are going to separate thumbnail and content files in future it can work

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -89,6 +90,18 @@ func (change *AllocationChange) Save(ctx context.Context) error {
 	return db.Save(change).Error
 }
 
+func (change *AllocationChange) GetAffectedFilePath() (string, error) {
```
dabasov: usually want to name funcs starting with Get only in case they return data directly, if they do some other actions we used to name them accordingly. I suggest to call it `ParseAffectedFilePath`, we can store path in change struct after the parse and have func `GetOrParseAffectedFilePath` to avoid multiple parses

code/go/0chain.net/core/common/connection.go
```diff
@@ -0,0 +1,72 @@
...
+		if diff >= ConnectionObjTimeout {
+			delete(connectionObjSizeMap, connectionID)
+		}
+	}
+}
+
+func init() {
+	go func() {
+		for {
+			time.Sleep(ConnectionObjCleanInterval)
```
dabasov: use ticker instead

code/go/0chain.net/core/lock/lock.go
```diff
@@ -37,6 +38,21 @@ func (m *Mutex) Unlock() {
 	m.mu.Unlock()
 }
 
+// TryUnlock unlock the lock if it is already locked otherwise do nothing.
+// Don't use tryUnlock and Unlock together for the same lock otherwise it may panic
+func (m *Mutex) TryUnlock() {
```
dabasov: we can remove it now, right?

dabasov (changes requested):

code/go/0chain.net/blobbercore/allocation/allocationchange.go
Roshan-Mehta: Modified.

code/go/0chain.net/core/common/connection.go
Roshan-Mehta: Modified

code/go/0chain.net/core/lock/lock.go
Roshan-Mehta: Yes, removed.

code/go/0chain.net/core/common/connection.go
```diff
@@ -0,0 +1,82 @@
+package common
```
lpoli: Lets not have this code in common package.
Its better to have it in allocation package

code/go/0chain.net/core/common/connection.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	// ConnectionObjCleanInterval start to clean the connectionObjMap
+	ConnectionObjCleanInterval = 10 * time.Minute
+	// ConnectionObjTimout after which connectionObj entry should be invalid
+	ConnectionObjTimeout = 10 * time.Minute
+)
+
+var (
+	connectionObjSizeMap = make(map[string]ConnectionObjSize)
+	connectionObjMutex   sync.Mutex
```
lpoli: It is going to be too many lock contentions. We should add benchmark test for this. For example have combinations of connection objects and allocation changes in order of over 10,000.

code/go/0chain.net/core/common/connection.go
```diff
@@ -0,0 +1,82 @@
...
+	if !ok {
+		return 0
+	}
+	return connectionObjSize.Size
+}
+
+// UpdateConnectionObjSize updates the connection size by addSize in memory
+func UpdateConnectionObjSize(connectionID string, addSize int64) {
+	connectionObjMutex.Lock()
+	defer connectionObjMutex.Unlock()
```
lpoli: Since these functions are short function, lets not use defer. Have it unlocked at the end.
Same for other functions

code/go/0chain.net/core/common/connection.go
```diff
@@ -0,0 +1,82 @@
...
+	Size      int64
+	UpdatedAt time.Time
+}
+
+// GetConnectionObjSize gets the connection size from the memory
+func GetConnectionObjSize(connectionID string) int64 {
+	connectionObjMutex.Lock()
+	defer connectionObjMutex.Unlock()
+	connectionObjSize, ok := connectionObjSizeMap[connectionID]
+	if !ok {
```
lpoli: No need to check this.
Directly ` returh connectionObjSizeMap[connectionID]` is better

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -39,6 +39,7 @@ const (
 	EncryptionHeaderSize = 128 + 128
 	// ReEncryptionHeaderSize re-encryption header size in chunk
 	ReEncryptionHeaderSize = 256
+	AllocationConnections  = "allocation_connections"
```
lpoli: Unused constant

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -607,7 +605,7 @@ func (fsh *StorageHandler) RenameObject(ctx context.Context, r *http.Request) (i
 	dfc := &allocation.RenameFileChange{ConnectionID: connectionObj.ID,
 		AllocationID: connectionObj.AllocationID, Path: objectRef.Path}
 	dfc.NewName = new_name
-	connectionObj.Size += allocationChange.Size
+	common.UpdateConnectionObjSize(connectionID, allocationChange.Size)
```
lpoli: No need to update size as it is zero.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -830,7 +822,7 @@ func (fsh *StorageHandler) MoveObject(ctx context.Context, r *http.Request) (int
 		DestPath:     destPath,
 	}
 	dfc.SrcPath = objectRef.Path
-	connectionObj.Size += allocationChange.Size
+	common.UpdateConnectionObjSize(connectionID, allocationChange.Size)
```
lpoli: No need to update size

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -952,15 +945,11 @@ func (fsh *StorageHandler) CreateDir(ctx context.Context, r *http.Request) (*blo
...
-	mutex := lock.GetMutex(connectionObj.TableName(), connectionID)
-	mutex.Lock()
-	defer mutex.Unlock()
-
 	allocationChange := &allocation.AllocationChange{}
 	allocationChange.ConnectionID = connectionObj.ID
 	allocationChange.Size = 0
 	allocationChange.Operation = constants.FileOperationCreateDir
-	connectionObj.Size += allocationChange.Size
+	common.UpdateConnectionObjSize(connectionID, allocationChange.Size)
```
lpoli: No need to update size

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -1041,25 +1030,12 @@ func (fsh *StorageHandler) WriteFile(ctx context.Context, r *http.Request) (*blo
 
 	elapsedRef := time.Since(st)
 	st = time.Now()
+
 	connectionObj, err := allocation.GetAllocationChanges(ctx, connectionID, allocationID, clientID)
 	if err != nil {
 		return nil, common.NewError("meta_error", "Error reading metadata for connection")
 	}
```
lpoli, line 1075: @Roshan-Mehta I don't see size update in connection object. Is it done somewhere for file upload?

code/go/0chain.net/blobbercore/handler/file_command_upload.go
```diff
@@ -134,18 +134,19 @@ func (cmd *UploadFileCommand) ProcessContent(ctx context.Context, req *http.Requ
...
+		common.UpdateConnectionObjSize(connectionObj.ID, fileOutputData.Size)
 	}
 
 	if allocationObj.BlobberSizeUsed+allocationSize > allocationObj.BlobberSize {
 		return result, common.NewError("max_allocation_size", "Max size reached for the allocation with this blobber")
 	}
 
 	cmd.fileChanger.AllocationID = allocationObj.ID
-	cmd.fileChanger.Size = allocationSize
+	cmd.fileChanger.Size += fileOutputData.Size
```
lpoli, line 145: Why did you modify this?

code/go/0chain.net/blobbercore/handler/file_command_upload.go
```diff
@@ -198,7 +200,8 @@ func (cmd *UploadFileCommand) reloadChange(connectionObj *allocation.AllocationC
 // UpdateChange replace AddFileChange in db
 func (cmd *UploadFileCommand) UpdateChange(ctx context.Context, connectionObj *allocation.AllocationChangeCollector) error {
 	for _, c := range connectionObj.Changes {
-		if c.Operation != constants.FileOperationInsert {
+		filePath, _ := c.GetOrParseAffectedFilePath()
+		if c.Operation != constants.FileOperationInsert || cmd.fileChanger.Path != filePath {
 			continue
 		}
 		c.Size = connectionObj.Size
```
lpoli, line 207: Does this still hold?

lpoli (changes requested):
You've done a good work.
A small suggestion:
Currently we check size against allocation capacity during commit only. We should also check this while Copying, Uploading/Updating. 
For example, for the allocation size of 1 GB client would upload 2 GB and will eventually fail in commit. But client occupied blobber's temp dir.

lpoli:
@Roshan-Mehta Also I don't see an endpoint to add a connection object. Please add and use it call from gosdk when multi-operation is happening

Roshan-Mehta:
> @Roshan-Mehta Also I don't see an endpoint to add a connection object. Please add and use it call from gosdk when multi-operation is happening

@lpoli  It will happen with the first operation. We don't need to add an additional endpoint for this. ConnectionObj and allocation creation will happen with the first operation itself. 

**Edit-1: As discussed offline, now the new endpoint for creating a connection object is added. It is cleaner way to do.**  @lpoli

code/go/0chain.net/blobbercore/handler/file_command_upload.go
Roshan-Mehta: It is no longer true. Earlier, there was only one file but now we have more files, the correct size of the file will no longer be equal to total allocation size change.

code/go/0chain.net/blobbercore/handler/file_command_upload.go
Roshan-Mehta: Yes with `cmd.fileChanger.Path != filePath` check.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: Removed

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: Removed

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: Removed

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: Removed

code/go/0chain.net/core/common/connection.go
Roshan-Mehta: Done

code/go/0chain.net/core/common/connection.go
Roshan-Mehta: Removed

Roshan-Mehta:
> Currently we check size against allocation capacity during commit only. We should also check this while Copying, Uploading/Updating.
> For example, for the allocation size of 1 GB client would upload 2 GB and will eventually fail in commit. But client occupied blobber's temp dir.

We are already doing this for upload and update.

Hitenjain14:
> > @Roshan-Mehta Also I don't see an endpoint to add a connection object. Please add and use it call from gosdk when multi-operation is happening
> 
> @lpoli It will happen with the first operation. We don't need to add an additional endpoint for this. ConnectionObj and allocation creation will happen with the first operation itself.

@Roshan-Mehta gosdk will send all the changes in parallel no? If yes there will be no such thing as first operation, all changes will try to create a new connectionObj.

Roshan-Mehta:
> @Roshan-Mehta gosdk will send all the changes in parallel no? If yes there will be no such thing as first operation, all changes will try to create a new connectionObj.

No, Gosdk sends the first operation and then other operations parallelly.
[[gosdk](https://github.com/0chain/gosdk/blob/7422acd7f644e6b78fe51d8c82bbae667d9c2182/zboxcore/sdk/multi_operation_worker.go#L53)]

code/go/0chain.net/blobbercore/allocation/connection.go
```diff
@@ -0,0 +1,78 @@
...
+
+// GetConnectionObjSize gets the connection size from the memory
+func GetConnectionObjSize(connectionID string) int64 {
+	connectionObjMutex.Lock()
+	defer connectionObjMutex.Unlock()
+	return connectionObjSizeMap[connectionID].Size
+}
+
+// UpdateConnectionObjSize updates the connection size by addSize in memory
+func UpdateConnectionObjSize(connectionID string, addSize int64) {
```
Hitenjain14, line 34: Can we have a map of pointers, we are creating a copy for every change.

code/go/0chain.net/blobbercore/allocation/connection.go
```diff
@@ -0,0 +1,78 @@
...
+	connectionObjMutex.Lock()
+	connectionObjSize := connectionObjSizeMap[connectionID]
+	connectionObjSizeMap[connectionID] = ConnectionObjSize{Size: connectionObjSize.Size + addSize, UpdatedAt: time.Now()}
```
Hitenjain14: Can we have a map of pointers, we are creating a copy for every change

lpoli: @Roshan-Mehta

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -520,7 +519,7 @@ func (fsh *StorageHandler) CommitWrite(ctx context.Context, r *http.Request) (*b
...
 
 	db.Delete(connectionObj)
-	common.DeleteConnectionObjEntry(connectionID)
+	allocation.DeleteConnectionObjEntry(connectionID)
```
lpoli: Since there can be many lock queues it might be better to run it in goroutine and avoid making client to wait.

lpoli (changes requested):
@Roshan-Mehta Please check for few comments and don't forget to add benchmark test

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Roshan-Mehta: Done

code/go/0chain.net/blobbercore/allocation/connection.go
Roshan-Mehta: Done

Hitenjain14 (approved):

Roshan-Mehta:
> > @Roshan-Mehta gosdk will send all the changes in parallel no? If yes there will be no such thing as first operation, all changes will try to create a new connectionObj.
> 
> No, Gosdk sends the first operation and then other operations parallelly. [[gosdk](https://github.com/0chain/gosdk/blob/7422acd7f644e6b78fe51d8c82bbae667d9c2182/zboxcore/sdk/multi_operation_worker.go#L53)]

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | test/multi-operation  |
| gosdk  | feature/multi_operation  |
| zboxcli  | feature/parallelizing-sync-command  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | test/multi-operation  |
| gosdk  | feature/multi_operation  |
| zboxcli  | feature/parallelizing-sync-command  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | test/multi-operation  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | feature/parallelizing-sync-command  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | multi_operation  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

dabasov (approved):
lgtm

PR #1084, opened 2023-05-15
Single save

### LINKED ISSUE #1073: Save changes atomically
All the changes should be saved to database in one transaction.

Hitenjain14:
### Tests

- 0chain:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/blobbercore/allocation/renamefilechange.go
```diff
@@ -41,11 +41,10 @@ func (rf *RenameFileChange) ApplyChange(ctx context.Context, change *AllocationC
 		return nil, common.NewError("invalid_reference_path", "file already exists")
 	}
 
-	affectedRef, err := reference.GetObjectTree(ctx, rf.AllocationID, rf.Path)
+	affectedRef, err := rootRef.GetSrcPath(rf.Path)
```
lpoli, line 44: This is not correct. You need to use GetObjectTree.

Hitenjain14: SrcPath function returns the same thing without making calls to DB, there could be a previous change which affected the children of srcRef and that change would have been saved in the memory.

lpoli: SrcPath is taken from reference path. It does not contain whole files that are affected by rename operation.

code/go/0chain.net/blobbercore/allocation/copyfilechange.go
```diff
@@ -37,12 +37,7 @@ func (rf *CopyFileChange) ApplyChange(ctx context.Context, change *AllocationCha
 			"maximum files and directories already reached: %v", err)
 	}
 
-	srcRef, err := reference.GetObjectTree(ctx, rf.AllocationID, rf.SrcPath)
-	if err != nil {
-		return nil, err
-	}
-
-	rootRef, err := reference.GetReferencePath(ctx, rf.AllocationID, rf.DestPath)
+	srcRef, err := rootRef.GetSrcPath(rf.SrcPath)
```
lpoli, line 40: srcRef does not contain object tree. We need to have objectTree. ObjectTree returns all the children recursively.

code/go/0chain.net/blobbercore/allocation/copyfilechange.go
```diff
@@ -92,19 +87,11 @@ func (rf *CopyFileChange) ApplyChange(ctx context.Context, change *AllocationCha
...
 		return nil, err
 	}
 
-	_, err = rootRef.CalculateHash(ctx, true)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, fileRef := range fileRefs {
-		reference.NewFileCreated(ctx, fileRef.ID)
```
lpoli, line 106: Why are we removing updates in stats table?

code/go/0chain.net/blobbercore/allocation/copyfilechange.go
```diff
@@ -160,3 +147,7 @@ func (rf *CopyFileChange) Unmarshal(input string) error {
 func (rf *CopyFileChange) CommitToFileStore(ctx context.Context) error {
 	return nil
 }
+
+func (rf *CopyFileChange) GetPath() []string {
+	return []string{rf.DestPath, rf.SrcPath}
```
lpoli, line 152: So here rf.srcPath is included. Maybe you are assuming it returns object Tree but it does not. Also the affected path is only destPath not srcPath.

Maybe its better to change function signature to `GetPath() string`

code/go/0chain.net/blobbercore/allocation/deletefilechange.go
```diff
@@ -107,14 +103,14 @@ func (nf *DeleteFileChange) CommitToFileStore(ctx context.Context) error {
...
-
-					if res.ThumbnailHash != "" {
-						err := filestore.GetFileStore().DeleteFile(nf.AllocationID, res.ThumbnailHash)
-						if err != nil {
-							logging.Logger.Error(fmt.Sprintf("Error while deleting thumbnail: %s", err.Error()),
-								zap.String("thumbnail", res.ThumbnailHash))
-						}
-					}
+					// We don't increase alloc size for thumbnail so we don't need to decrease it
+					// if res.ThumbnailHash != "" {
```
lpoli, line 107: Are we deleting it elsewhere?

code/go/0chain.net/blobbercore/reference/object.go
```diff
@@ -29,12 +27,13 @@ func DeleteObject(ctx context.Context, allocationID, objPath string, ts common.T
 	if err != nil {
 		return nil, err
 	}
-
-	parentPath := filepath.Dir(objPath)
-	rootRef, err := GetReferencePath(ctx, allocationID, parentPath)
-	if err != nil {
-		return nil, err
+	if objPath == "/" {
```
lpoli, line 30: @Hitenjain14 Consider a case in list of changes where we delete root path and and upload file. If delete root path change is applied after upload file change then allocation root is root of empty allocation. Otherwise it is of uploaded file.
Order of changes matters here. Maybe we should restrict conflicting changes or let client send order of changes while committing.

code/go/0chain.net/blobbercore/allocation/movefilechange.go
```diff
@@ -22,17 +22,10 @@ func (rf *MoveFileChange) DeleteTempFile() error {
...
 
-	srcRef, err := reference.GetObjectTree(ctx, rf.AllocationID, rf.SrcPath)
-	if err != nil {
-		return nil, err
-	}
-
-	rootRef, err := reference.GetReferenceForHashCalculationFromPaths(ctx, rf.AllocationID,
-		[]string{rf.DestPath, filepath.Dir(rf.SrcPath)})
-
+	srcRef, err := rootRef.GetSrcPath(rf.SrcPath)
```
lpoli, line 28: Use object tree

code/go/0chain.net/blobbercore/allocation/movefilechange.go
```diff
@@ -165,3 +155,7 @@ func (rf *MoveFileChange) Unmarshal(input string) error {
 func (rf *MoveFileChange) CommitToFileStore(ctx context.Context) error {
 	return nil
 }
+
+func (rf *MoveFileChange) GetPath() []string {
+	return []string{rf.DestPath, filepath.Dir(rf.SrcPath)}
```
lpoli: See comment in copyfilechange.go

code/go/0chain.net/blobbercore/allocation/movefilechange.go
```diff
@@ -105,21 +98,18 @@ func (rf *MoveFileChange) ApplyChange(ctx context.Context, change *AllocationCha
...
-	}
 
 	for _, fileRef := range fileRefs {
 		fileRef.IsPrecommit = true
 	}
-	return rootRef, err
+	return rootRef, nil
 }
 
 func (rf *MoveFileChange) processCopyRefs(
```
lpoli: Lets rename this function to processMoveRefs

code/go/0chain.net/blobbercore/allocation/file_changer_update.go
```diff
@@ -81,9 +76,6 @@ func (nf *UpdateFileChanger) ApplyChange(ctx context.Context, change *Allocation
 	fileRef.HashToBeComputed = true
 	nf.deleteHash = make(map[string]int)
 
-	if fileRef.ThumbnailHash != "" && fileRef.ThumbnailHash != nf.ThumbnailHash {
```
lpoli, line 84: Are we removing thumbnail at all? If so its fine.

code/go/0chain.net/blobbercore/allocation/deletefilechange.go
```diff
@@ -30,18 +30,14 @@ type DeleteFileChange struct {
 	Hash         string `json:"hash"`
 }
 
-func (nf *DeleteFileChange) ApplyChange(ctx context.Context, change *AllocationChange,
+func (nf *DeleteFileChange) ApplyChange(ctx context.Context, rootRef *reference.Ref, change *AllocationChange,
 	allocationRoot string, ts common.Timestamp, _ map[string]string) (*reference.Ref, error) {
 
-	rootRef, err := reference.DeleteObject(ctx, nf.AllocationID, filepath.Clean(nf.Path), ts)
+	_, err := reference.DeleteObject(ctx, rootRef, nf.AllocationID, filepath.Clean(nf.Path), ts)
```
lpoli: We send pointer to deleteobject function. It makes no sense to return same pointer. Lets remove it in DeleteObject function

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -557,49 +550,27 @@ func (r *Ref) SaveDirRef(ctx context.Context) error {
...
-			}
-			r.ID = 0
-			r.IsPrecommit = true
-			err = tx.Create(r).Error
-			if err != nil {
-				return err
-			}
-			return nil
-		})
+		err := db.Delete(&Ref{}, "id=?", r.ID).Error
```
lpoli, line 555: Is this delete related to two-phase commit?

lpoli (changes requested):

code/go/0chain.net/blobbercore/reference/object.go
Hitenjain14: We had already discussed that there won't be any conflicting changes sent in a single commit

code/go/0chain.net/blobbercore/allocation/deletefilechange.go
Hitenjain14: We delete it in MoveFileStore in the next commit

code/go/0chain.net/blobbercore/allocation/copyfilechange.go
Hitenjain14: As we create a new row in the table on every update and soft delete previous one, and we will need new ID to update stats.

code/go/0chain.net/blobbercore/allocation/file_changer_update.go
Hitenjain14: No we are not, but we don't increase the number of files or file size when thumbnail is uploaded so we don't delete it here as well. We delete this from filestore on next commit.

code/go/0chain.net/blobbercore/reference/ref.go
Hitenjain14: Yes we soft delete the previous row and create a new one to have 2 versions.

code/go/0chain.net/blobbercore/allocation/renamefilechange.go
dabasov: this object tree should have been already loaded before, we can pass reference to it, or convert it to scrPath

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -786,6 +787,10 @@ func (fsh *StorageHandler) CopyObject(ctx context.Context, r *http.Request) (int
 	if err != nil {
 		return nil, common.NewError("invalid_parameters", "Invalid file path. "+err.Error())
 	}
+	dirPath := filepath.Dir(objectRef.Path)
+	if dirPath == destPath || objectRef.Path == destPath {
```
lpoli: objectRef.ParentPath instead of filepath.Dir(objectRef.Path)

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
```diff
@@ -893,6 +898,10 @@ func (fsh *StorageHandler) MoveObject(ctx context.Context, r *http.Request) (int
 	if err != nil {
 		return nil, common.NewError("invalid_parameters", "Invalid file path. "+err.Error())
 	}
+
+	if objectRef.ParentPath == destPath {
```
lpoli, line 932: This check is duplicate

code/go/0chain.net/blobbercore/handler/storage_handler.go
```diff
@@ -537,7 +537,8 @@ func (fsh *StorageHandler) getReferencePath(ctx context.Context, r *http.Request
 		errCh <- common.NewError("invalid_signature", "could not verify the allocation owner or colloborator")
 		return
 	}
-	rootRef, err := reference.GetReferencePathFromPaths(ctx, allocationID, paths)
+	objTreePath := []string{}
```
lpoli: No need to initialize variable. Directly use []string{}

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -391,9 +396,13 @@ func deleteFromFileStore(ctx context.Context, allocationID string) error {
 
 func (a *AllocationChangeCollector) GetRootRef(ctx context.Context) (*reference.Ref, error) {
 	paths := make([]string, 0)
-
+	objTreePath := make([]string, 0)
 	for _, change := range a.AllocationChanges {
-		paths = append(paths, change.GetPath()...)
+		allPaths := change.GetPath()
+		paths = append(paths, allPaths...)
```
lpoli, line 403: Its better to append `allPaths[0]`. GetReferencePath will load first children of all paths it encounters and we don't need reference path for srcPath.
You need to use AddChild function to attach objectTree to rootRef.

code/go/0chain.net/blobbercore/reference/ref.go
```diff
@@ -474,6 +474,7 @@ func (r *Ref) AddChild(child *Ref) {
 	for i, ref := range r.Children {
 		if strings.Compare(child.Name, ref.Name) == 0 {
```
lpoli, line 476: Why do we compare with name and path in two steps?
We can compare with either of path and name and in single step. Also use the result instead of comparing it two places.

lpoli: Not a big deal but I don't think we ever require to add already existing child.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
```diff
@@ -265,6 +265,11 @@ func (a *AllocationChangeCollector) MoveToFilestore(ctx context.Context) error {
 
 			for _, ref := range refs {
 
+				var count int64
+				tx.Model(&reference.Ref{}).
+					Where("allocation_id=? AND validation_root=? AND type=?", a.AllocationID, ref.PrevValidationRoot, reference.FILE).
```
lpoli, line 270: I suppose its better to add isPreCommit condition.

lpoli (commented):
Please also add following if not already there:
1. Unit test for copying directory
2. Unit test for moving directory
3. Unit test for renaming directory

@Hitenjain14 Also I don't see any code where we are restricting conflicting changes. You can add it before Applying changes.

code/go/0chain.net/blobbercore/reference/ref.go
Hitenjain14: There was an issue with nested move and copy 

> Please also add following if not already there:
> 
> 1. Unit test for copying directory
> 2. Unit test for moving directory
> 3. Unit test for renaming directory
> 
> @Hitenjain14 Also I don't see any code where we are restricting conflicting changes. You can add it before Applying changes.

The client is responsible for that.

Hitenjain14:
> Please also add following if not already there:
> 
> 1. Unit test for copying directory
> 2. Unit test for moving directory
> 3. Unit test for renaming directory
> 
> @Hitenjain14 Also I don't see any code where we are restricting conflicting changes. You can add it before Applying changes.

The client is responsible to not send conflicting changes in single commit.
Unit tests are added for all operations including multi operations.

code/go/0chain.net/blobbercore/reference/ref.go
Hitenjain14: We don't but we were doing it in processCopy and Move before, to solve that this has been done https://github.com/0chain/blobber/pull/1088

code/go/0chain.net/blobbercore/handler/storage_handler.go
Hitenjain14: I kept it as I thought gosdk will send us the paths, but will remove it as its not needed anymore.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
Hitenjain14: Its better to return if the parent path is same as destPath, then loading refs from db in the next step no?

code/go/0chain.net/blobbercore/allocation/allocationchange.go
Hitenjain14: We need referencePath for move operation.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
Hitenjain14: I don't think so its required there, we are just checking if this prevRoot is not the current root of any other Ref, be it preCommit or not.

code/go/0chain.net/blobbercore/handler/object_operation_handler.go
lpoli: I mean you have already done this check above.

code/go/0chain.net/blobbercore/allocation/allocationchange.go
lpoli: I see, as both path are affected. So lets append only for move.
I am emphasizing it because it is costly query and operation

Hitenjain14: How to attach the objectTree to rootRef, AddChild adds the path in sorted order. If we don't want to do this for copy then I guess we should add one more field to the ApplyChange function.

dabasov (approved):
lgtm

lpoli (approved):
I see that you've fixed nil proof issue. That's fine.
For the comment part, please add some details. It is not enough for future developer to understand the subtleties.

